PR #12, opened 2019-08-02
Fix find all algos

JacobPlaster:
Currently is seems like the lowdb `find` validation is not working properly. As this is causing some big problems this pr hot fixes that by using `getAll` and mapping to validate active orders.


JacobPlaster:
fyi @f3rno

PR #14, opened 2019-08-02
Update/unregister ao

AntonNazarenko:


lib/host/ui/register_ao_uis.js
```diff
@@ -10,7 +10,7 @@ const { PLATFORM = 'bitfinex' } = process.env
  *
  * @param {Object} aoHost - algo order host
  */
-module.exports = async (aoHost = {}) => {
+module.exports = async (aoHost = {}, label = null) => {
```
JacobPlaster: What is label here?

AntonNazarenko: label of a AO: 'Ping/pong', 'Iceberg' etc.

lib/host/ui/unregister_ao_uis.js
```diff
@@ -0,0 +1,51 @@
...
+const _isFunction = require('lodash/isFunction')
+
+const { PLATFORM = 'bitfinex' } = process.env
+
+/**
+ * Registers all order form layouts for the provided AO Host
+ *
+ * @param {Object} aoHost - algo order host
+ */
+module.exports = async (aoHost = {}, aoID) => {
```
JacobPlaster: Hmm im not sure if this is the best way to handle this, maybe we should introduce an 'Active' flag similar to the AOs? The problem with this is that there isnt any state of the definition being de-registered. So if the AOs are re-registed to the UI via (https://github.com/bitfinexcom/bfx-hf-algo/blob/master/lib/host/ui/register_ao_uis.js) then previously de-registered AO will appear in the UI again

JacobPlaster: This might cause a problem in the future when we start allowing the user to create new AOs and therefore promp a call to `register_ao_uis`

AntonNazarenko: Sorry, we need to delete this file, it's unused

lib/host/ui/register_ao_uis.js
JacobPlaster: Right got you, I dont this is the best solution either as this means we can only ever disable one order at a time  and it also creates a bit of a confusing side effect. Leave this with me, im going to do a spike to see if we can add this logic to the hf-server instead, Im thinking we could totally destroy and re-create the hf-algo instance when a user re-registers. This way, not only will it be re-registered from the UI but it also wont be able to perform any updates

vigan-abd:
outdated

Issue #15, opened 2019-08-08
Noce code change in bfx-api-node-util has broken order creation


JacobPlaster:
This pulll request https://github.com/bitfinexcom/bfx-api-node-util/commit/9e0d87f4965f518d23b4a55ef7249e95713e6c14 changed the way the nonce is generated.
THe problem is that all of the default order examples use this function to also generate cId when creating an order which is now invalid due to the `* 1000` multiplier

robertkowalski:
fixed

PR #16, opened 2019-08-08
Use date.now as cID instead of nonce()

JacobPlaster:
Fixes issue https://github.com/bitfinexcom/bfx-hf-algo/issues/15


JacobPlaster:
Manually tested and all is good

PR #22, opened 2019-12-03
2.0.5 Feature: Derivatives Support

f3rno:
### Description:
Enables futures support for all algo orders (filling in the `lev` field & setting the order type correctly)

### New features:
- [x] futures support

### PR status:
- [x] Version bumped
- [x] Change-log updated
- [ ] Tests added or updated
- [ ] Documentation updated

### DIFF
--- a/lib/iceberg/meta/validate_params.js
+++ b/lib/iceberg/meta/validate_params.js
@@ -5,7 +5,8 @@ const _isFinite = require('lodash/isFinite')
 
 module.exports = (args = {}) => {
   const {
-    price, amount, sliceAmount, orderType, submitDelay, cancelDelay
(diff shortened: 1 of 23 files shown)


lib/iceberg/meta/validate_params.js
```diff
@@ -24,5 +25,11 @@ module.exports = (args = {}) => {
     return 'Amount & slice amount must have same sign'
   }
 
+  if (_futures) {
+    if (!_isFinite(lev)) return 'Invalid leverage'
+    if (lev < 1) return 'Leverage less than 1'
+    if (lev > 100) return 'Leverage greater than 100'
```
JacobPlaster: Might want to make this more customization since BFX may want to offer more than 100 in the future. Also Binance offer 125x lev (which is absolutely insane btw...)

PR #41, opened 2020-11-17
test: fix/improve testsuite

robertkowalski:
1. an event emitter is always async
2. the timeout is redundant

### DIFF
--- a/test/lib/host/gen_helpers.js
+++ b/test/lib/host/gen_helpers.js
@@ -73,21 +73,13 @@ describe('genHelpers', () => {
     it('emits on the self scope after a timeout using the helper state', (done) => {
       const ev = new AsyncEventEmitter()
       const h = H({ ev })
-      let eventSeen = false
 
       ev.once('self:test', (...args) => {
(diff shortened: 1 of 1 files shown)


test/lib/host/gen_helpers.js
```diff
@@ -123,7 +123,7 @@ describe('genHelpers', () => {
       setTimeout(() => {
         assert.ok(eventSeen, 'did not emit test event')
         done()
-      }, 1)
+      }, 20)
```
Daniel1984, line 126: feels like `eventSeen` in both cases is redundant and `done()` callback can be called from within `ev.once('self:test', (...args) => {` no?

Daniel1984 (approved):
small comment but other than that lgtm

test/lib/host/gen_helpers.js
robertkowalski: true, the test itself is a bit off - a event emitter is always async, and the timeout thing is not making a difference

vigan-abd (approved):

PR #42, opened 2020-11-17
iceberg: fix float issues

robertkowalski:


Daniel1984 (approved):

vigan-abd (approved):
looks good, but maybe we could use https://github.com/bitfinexcom/lib-js-util-math instead of raw bignumber.js since it adds more validation against NaN/null values through nBN?

vigan-abd (approved):

PR #46, opened 2020-12-08
ping/pong: fix forms when ordercount was changed

robertkowalski:
steps to reproduce:

1. open the ping/pong order form
2. in order count, set value to 2
3. set it back to 1
4. there is no price field any more
5. resulting order after a submit has a price of '0'

the ui submits always strings, and the `eq` operator in the
form framework uses a strict equal comparision.

fixes: https://github.com/bitfinexcom/bfx-hf-ui/issues/171


Daniel1984 (approved):
🌮

vigan-abd (approved):

PR #48, opened 2021-01-05
allow post-only limit orders in accumulate/distribute algo

avsek477:
### Description:
added post only orders for accumulate/distribute algorithmic orders

### New features:
- [ postonly orders option available for accumulate/distribute algo]

### DIFF
--- a/lib/accumulate_distribute/meta/validate_params.js
+++ b/lib/accumulate_distribute/meta/validate_params.js
@@ -50,7 +50,7 @@ const validateParams = (args = {}, pairConfig = {}) => {
   const {
     limitPrice, amount, sliceAmount, orderType,
     intervalDistortion, amountDistortion, sliceInterval, relativeOffset,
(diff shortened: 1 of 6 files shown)


lib/accumulate_distribute/meta/validate_params.js
```diff
@@ -77,6 +77,10 @@ const validateParams = (args = {}, pairConfig = {}) => {
     return validationErrObj('limitPrice', 'Limit price required for LIMIT order type')
   }
 
+  if (orderType === 'LIMIT' && !_isBoolean(postonly)) {
```
vigan-abd, line 80: should postonly be always true in case of limit order?

vigan-abd (approved):
one questions, looks good

lib/accumulate_distribute/meta/validate_params.js
avsek477: nope.. it can be false as well .. its up to the users on how they wanna submit the order

PR #49, opened 2021-01-15
Allow one requeue

avsek477:


lib/host/events/resubmit_all_orders.js
```diff
@@ -0,0 +1,46 @@
...
+ * @returns {object} nextInstanceState
+ */
+module.exports = async (aoHost, gid, orders, delay) => {
+  await withAOUpdate(aoHost, gid, async (instance = {}) => {
+    const { state = {}, h = {} } = instance
+    const { id } = state
+    const { submitOrderWithDelay, debug } = h
+    const _orders = _isObject(orders)
+      ? Object.values(orders).filter(order => order.status === 'REQUEUE')
+      : orders.filter(order => order.status === 'REQUEUE')
```
robertkowalski, line 23: odd that the internal format for the data changes sometimes... its sometimes an object and sometimes an array?

avsek477: hmm.. depends on the format you send it.. its just like a precaution so that it works for both cases.. there are formats where you save the orders in key value format where key is the ocid assigned to the order and value being the order object.. otherwise the orders are sent directly in the array format

PR #50, opened 2021-01-15
docs: create user docs for non-coders 

robertkowalski:
with people coming from all walks of life using algo orders, using
JSdoc to explain algo orders does not work well. very often the behaviour
of an algo order is unclear to users who have never heard of jsdoc.

this PR takes the current jsdoc summaries as a starting base
for more user oriented documentation to serve a larger user base.

additionally we extend ma/crossover orders with general informations on data collection and runtime behaviour


user-docs.md
```diff
@@ -0,0 +1,99 @@
...
+component fills before submitting subsequent orders. Enabling the 'Catch Up'
+flag will cause the algorithm to ignore the slice interval for the next order
+if previous orders have taken longer than expected to fill, thereby ensuring
+the time-to-fill for the entire order is not adversely affected.
+
+The price must be manually specified as `limitPrice` for `LIMIT` order types,
+or as a combination of a price offset & cap for `RELATIVE` order types.
+`MARKET` A/D orders execute using `MARKET` atomic orders, and offer no price
+control.
+
```
avsek477, line 30: could you add this as well here ?
 * Post-Only limit orders can also be submitted. However, when 'Await Fill' option
 * is enabled, if a sliced postonly limit order is cancelled, the algorithm will 
 * wait for the the cancelled order to fill before submitting other subsequent orders.
 * Otherwise, the algorithm just cancels the postonly orders.

user-docs.md
```diff
@@ -0,0 +1,99 @@
...
+ - [Accumulate/Distribute](#ad)
+ - [Iceberg](#iceberg)
+ - [MA Crossover](#macross)
+ - [Order Creates OCO](#ococo)
+ - [Ping/Pong](#pingpong)
+ - [TWAP](#twap)
+
+## Accumulate/Distribute
+
+<a id="ad" />
```
avsek477, line 14: is id is required?

user-docs.md
robertkowalski: yes, good idea! lets add it once the feature is merged

PR #51, opened 2021-01-19
added random fuzziness to twap amount

avsek477:


lib/twap/util/generate_order.js
```diff
@@ -20,13 +22,22 @@ const { DUST } = Config
...
   const {
-    sliceAmount, orderType, symbol, amount, lev, _margin, _futures
+    sliceAmount, orderType, symbol, amount, lev, _margin, _futures, amountDistortion
   } = args
+  let orderAmount
 
+  if (_isFinite(amountDistortion)) {
+    const plusOrMinus = Math.random() > 0.5 ? 1 : -1
+    const randomPercentage = ((Math.random() * amountDistortion * plusOrMinus) + 1) // random percentage from 1 to amountDistortion
+    orderAmount = sliceAmount * randomPercentage
```
robertkowalski, line 32:
```
> 0.1 * 0.2
0.020000000000000004
```

can you use the bignumber lib we use also for iceberg orders?

lib/twap/util/generate_order.js
```diff
@@ -1,6 +1,8 @@
 'use strict'
 
+const _isFinite = require('lodash/isFinite')
```
robertkowalski, line 3: can you add a unit test for `generate_order.js`? should also cover the float issue

lib/twap/util/generate_order.js
avsek477: i used the prepareAmount from bfx-api-node-util library for it .. so thought not using bignumber would be okay .. but sure i will use bignumber for this

lib/twap/util/generate_order.js
```diff
@@ -20,13 +22,22 @@ const { DUST } = Config
...
   const { args = {}, remainingAmount } = state
   const {
-    sliceAmount, orderType, symbol, amount, lev, _margin, _futures
+    sliceAmount, orderType, symbol, amount, lev, _margin, _futures, amountDistortion
   } = args
+  let orderAmount
 
+  if (_isFinite(amountDistortion)) {
+    const plusOrMinus = Math.random() > 0.5 ? 1 : -1
+    const randomPercentage = ((Math.random() * amountDistortion * plusOrMinus) + 1) // random percentage from 1 to amountDistortion
```
robertkowalski, line 31: just to double check, lets say i choose a 2% distortion.

i have an initial value of 100 - so the submtited values are withign the range of 98 - 102, correct?

can you describe and test if everything works with the overall order amount?

e.g. i have an overall amount of 100, distortion 2%, with 5 slices...

does the algo then could overuse amounts from adding up additional distortions?

robertkowalski (changes requested):
nice progress so far! needs some more testing

lib/twap/util/generate_order.js
avsek477: the submitted orders never exceeds the total amount entered by the user despite the additional distortions.. in every interval tick, if tradeBeyondEnd is enabled, the algo ensures that the next order would not push us over the total amount and doesnt submit new orders; and when tradeBeyondEnd is disabled, the remaining amount is sent for the new order when the next order pushes over the total amount 
i will write the test cases covering these possible cases

lib/twap/util/generate_order.js
```diff
@@ -18,16 +21,25 @@ const { DUST } = Config
  * @returns {object} order - null if no amount remains
  */
 const generateOrder = (state = {}, price) => {
-  const { args = {}, remainingAmount } = state
+  const { orders = {}, args = {}, remainingAmount } = state
   const {
-    sliceAmount, orderType, symbol, amount, lev, _margin, _futures
+    sliceAmount, orderType, symbol, amount, lev, _margin, _futures, amountDistortion
   } = args
+  let orderAmount=0
```
robertkowalski, line 28: style: `orderAmount = 0`

lib/twap/meta/get_ui_def.js
```diff
@@ -105,6 +106,12 @@ const getUIDef = () => ({
       priceField: 'price'
     },
 
+    amountDistortion: {
+      component: 'input.percent',
+      label: 'Amount Distortion %',
+      customHelp: 'Amount to distort individual order sizes to prevent detection, in percent'
```
robertkowalski, line 112: default should be 0

PR #52, opened 2021-01-21
twap: improve submit delay

robertkowalski:
with the current submit delay the algo takes  the price of an
orderbook first, then creates an order based on that price and
THEN waits for a default of 2 seconds before submitting it.
that means, when the order submit delay is triggered, an outdated
orderbook price was taken.

this PR applies the timeout before the price is set, providing
better prices that are more up to date.

it also reduces issues where an unfilled TWAP order was cancelled,
but was sitting still on the book, while the price for the
next order was already decided upon, leading the algo trading
against itself, constantly lowering or increasing the price by
calculating the mid-price too early. this problem will be fully
addressed in another PR

### DIFF
--- a/lib/host/gen_helpers.js
+++ b/lib/host/gen_helpers.js
@@ -91,6 +91,12 @@ const genHelpers = (state = {}, adapter) => {
       pendingTimeouts = []
     },
 
+    timeout: (ms) => {
+      return new Promise((resolve) => {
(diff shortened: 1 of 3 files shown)


vigan-abd (approved):
lgtm

avsek477 (approved):
lgtm

vigan-abd (approved):

PR #57, opened 2021-02-16
float issue fixes for icerberg algo

avsek477:
Minor fix for modifying the slice amount while processing the parameters for iceberg orders.


robertkowalski (approved):
good catch, lgtm

PR #59, opened 2021-02-16
float issue fix for twap algo

avsek477:
Using bignumber.js to avoid float issue while calculating the order amounts.

### DIFF
--- a/test/lib/twap/events/self_interval_tick.js
+++ b/test/lib/twap/events/self_interval_tick.js
@@ -24,6 +24,32 @@ const timeout = () => {
 }
 
 describe('twap:events:self_interval_tick', () => {
+  it('submits order for float order amount', async () => {
+    let orderSubmitted = false
+    const instance = {
+      state: {
+        gid: 100,
+        orders: { o: { amount: 0.1 }, p: { amount: 0.1 } },
+        args: {
+          ...args,
+          amount: 0.3,
+          orderType: 'MARKET'
+        }
+      },
+      h: {
+        timeout,
+        debug: () => {},
+        emit: (eventName) => {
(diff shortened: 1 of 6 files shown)


robertkowalski (changes requested):
this requires tests for the changes which are failing on current master and are fixed with the patch

robertkowalski:
`gen_preview.js` missing

test/lib/twap/util/is_target_met.js
```diff
@@ -6,8 +6,8 @@ const isTargetMet = require('../../../../lib/twap/util/is_target_met')
 
 describe('twap:util:is_target_met', () => {
   it('verifies that the price target is met', () => {
-    assert.ok(isTargetMet({ priceTarget: 1, priceDelta: 0 }, 1))
-    assert.ok(isTargetMet({ priceTarget: 1, priceDelta: 0.2 }, 1.2))
-    assert.ok(!isTargetMet({ priceTarget: 1, priceDelta: 0.2 }, 1.3))
+    assert.ok(isTargetMet({ priceTarget: 5.6, priceDelta: 0 }, 5.6))
```
robertkowalski, line 9: in the best case an additional test case is added, so its clear that it also handles floats and people do not change the test in the future

test/lib/twap/events/self_interval_tick.js
```diff
@@ -24,6 +24,32 @@ const timeout = () => {
 }
 
 describe('twap:events:self_interval_tick', () => {
+  it('submits order for float order amount', async () => {
```
robertkowalski, line 27: nice one!

robertkowalski (approved):
lgtm

PR #60, opened 2021-02-18
test cases for generating ping pong table

avsek477:
Test case to check the ping and pong prices for generated ping pong table.


robertkowalski (approved):
lgtm

PR #62, opened 2021-02-23
clearing interval while removing algo order instances

avsek477:
On meta reload, the instances of the algo orders are removed and later created as required when the users choose to reload the orders manually. However, the interval set for these instances aren't cleared while removing the instances causing a memory leak. This PR clears the interval for these active instances while removing those instances of algo orders.


robertkowalski (approved):
lgtm

PR #63, opened 2021-03-04
single atomic cancellation while resuming orders issue fixed

avsek477:
While resuming the algo orders, the format of the received snapshot of active orders (when only one atomic order is active) wasn't as per the written logic for cancelling the active atomic orders. This is a fix to handle cancellation of active atomic order before resuming the algo order in order to prevent the same atomic order to be submitted twice.


robertkowalski (approved):
lgtm

PR #64, opened 2021-03-04
prevent initial order to be submitted twice in oco order 

avsek477:
While resuming the oco algo order, we weren't keeping track if the initial submitted order was filled or not causing the algo to submit the same order twice. This fix checks if the initial order was filled or not and based on this, decides to resume with the required orders that needs to be submitted.


robertkowalski (approved):
lgtm

PR #65, opened 2021-03-04
twap: fix distort amount calc 

robertkowalski:
don't calculate when the distortion is set to 0 (default value)


avsek477 (approved):
the +1 in randomPercentage handles this case but its better this way than to calculate using binumber when not needed. lgtm :+1:

PR #66, opened 2021-03-05
reset priceTarget  and priceCondition for market orders

avsek477:
In the UI,  the value of priceTarget and priceCondition isn't reset when user changes order type from limit to market. This is a fix to change these values as per required in case of market order type.


robertkowalski (approved):
lgtm

PR #70, opened 2021-03-05
twap: fix interval queing up

robertkowalski:
when we cancel an order, we want to wait until its removed from
the book so we don't trade against our own order.

we also don't want to queue up operations.

currently a simple interval is used for scheduling a twap algo run.
i.e. every two seconds the algo starts, checks the orderbook,
cancels and submits an order. both sending and cancelling an order
is async. that means with a slow connection it can take longer
than the interval. that can lead to queing up operations.

another case is when the order book is outdated. that happens,
with a slow connection especially when the algo starts to run and
new subscription data has not arrived yet. because we were not
waiting for async cancel/submit to be finished, they woudl queue up,
whil ein the background new intervals would already kick off new runs
in parallel. that leads to an issue were the twap algo trades against
its own outdated orders in the orderbook.

the new approach waits for all operations of the current tick to
get finished, before allowing a new run, avoiding race conditions
and wrong prices

### DIFF
--- a/lib/host/gen_helpers.js
+++ b/lib/host/gen_helpers.js
@@ -92,9 +92,13 @@ const genHelpers = (state = {}, adapter) => {
     },
 
     timeout: (ms) => {
-      return new Promise((resolve) => {
-        setTimeout(resolve, ms)
+      let id = null
+
+      const p = new Promise((resolve) => {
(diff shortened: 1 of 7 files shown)


avsek477 (approved):
lgtm :+1:

PR #72, opened 2021-03-08
disable the adjacent option on user input between slice amount and sl…

avsek477:
Disable the adjacent option when user inputs value for either slice amount or slice percent.
Related PR: https://github.com/bitfinexcom/bfx-hf-ui/pull/298


robertkowalski (approved):
lgtm

PR #74, opened 2021-03-08
change slice amount tooltip and validate if its greater than the tota…

avsek477:
1. Change the slice amount tooltip.
2. Condition check if slice amount is greater than  than total amount.


robertkowalski (approved):
lgtm

PR #77, opened 2021-03-12
cancel all orders set by acc/dis algo order on life stop event

avsek477:
This removes all of the respective atomic orders set by acc/dis algo when stopped by the user.


robertkowalski (approved):
lgtm

PR #78, opened 2021-03-12
cancels all orders set by ma crossover algo on life_stop event

avsek477:
This removes all of the respective atomic orders set by ma crossover algo when stopped by the user.

### DIFF
--- a/lib/ma_crossover/events/life_stop.js
+++ b/lib/ma_crossover/events/life_stop.js
@@ -8,6 +8,14 @@
  * @param {AOInstance} instance - AO instance
  * @returns {Promise} p - resolves on completion
  */
-const onLifeStop = async (instance = {}) => {}
+const onLifeStop = async (instance = {}) => {
+  const { state = {}, h = {} } = instance
+  const { orders = {}, gid } = state
+  const { emit, debug } = h
+
+  debug('detected ma crossover algo cancelation, stopping...')
+
+  await emit('exec:order:cancel:all', gid, orders)
+}
 
 module.exports = onLifeStop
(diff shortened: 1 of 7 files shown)


lib/ma_crossover/events/life_stop.js
```diff
@@ -8,6 +8,14 @@
...
  */
-const onLifeStop = async (instance = {}) => {}
+const onLifeStop = async (instance = {}) => {
+  const { state = {}, h = {} } = instance
+  const { orders = {}, gid } = state
+  const { emit, debug } = h
+
+  debug('detected ma crossover algo cancelation, stopping...')
+
+  await emit('exec:order:cancel:all', gid, orders)
```
robertkowalski, line 18: take care with that, ma crossover is a special case...

robertkowalski: https://github.com/bitfinexcom/bfx-hf-algo/blob/79f857bb333387d38a13350bcc9709bc91abea9d/lib/ma_crossover/events/data_managed_candles.js#L92-L102

robertkowalski: when a cross happens, it will submit an order and exit with `exec:stop`. 

in case the user selected limit orders, we would cancel it

robertkowalski: maybe you can pass a parameter via the event system to `onLifeStop`

```js
return emit('exec:stop', { keepOrdersOpen: true })
```

avsek477: used this to keep orders alive instead of using `updateState`.. thanks!!

robertkowalski (approved):
lgtm

PR #79, opened 2021-03-12
cancels all orders set by ping pong algo on life_stop event

avsek477:
This removes all of the respective atomic orders set by ping pong algo when stopped by the user.

### DIFF
--- a/lib/ping_pong/events/life_stop.js
+++ b/lib/ping_pong/events/life_stop.js
@@ -12,7 +12,9 @@
 const onLifeStop = async (instance = {}) => {
   const { state = {}, h = {} } = instance
   const { orders = {}, gid } = state
-  const { emit } = h
+  const { emit, debug } = h
+
+  debug('detected ping/pong algo cancelation, stopping...')
 
   return emit('exec:order:cancel:all', gid, orders)
 }
--- a/lib/ping_pong/events/orders_order_cancel.js
+++ b/lib/ping_pong/events/orders_order_cancel.js
@@ -12,13 +12,11 @@
  * @returns {Promise} p - resolves on completion
  */
 const onOrdersOrderCancel = async (instance = {}, order) => {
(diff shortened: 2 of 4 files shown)


lib/ping_pong/events/life_stop.js
```diff
@@ -12,9 +12,11 @@
 const onLifeStop = async (instance = {}) => {
   const { state = {}, h = {} } = instance
   const { args = {}, orders = {}, gid } = state
-  const { emit } = h
+  const { emit, debug } = h
   const { cancelDelay } = args
 
+  debug('detected ping/pong algo cancelation, stopping...')
+
   return emit('exec:order:cancel:all', gid, orders, cancelDelay)
```
robertkowalski, line 20: should be set to 0 here, else we have ghost orders after the user clicks cancel

lib/ping_pong/events/orders_order_cancel.js
```diff
@@ -12,14 +12,11 @@
...
 const onOrdersOrderCancel = async (instance = {}, order) => {
-  const { state = {}, h = {} } = instance
-  const { args = {}, orders = {}, gid } = state
+  const { h = {} } = instance
   const { emit, debug } = h
-  const { cancelDelay } = args
 
   debug('detected atomic cancelation, stopping...')
 
-  await emit('exec:order:cancel:all', gid, orders, cancelDelay)
```
robertkowalski, line 22: here we can use cancelDelay, but i think we should remove that feature all together...

avsek477: agreed

lib/ping_pong/events/life_stop.js
avsek477: ok!! will change to 0

robertkowalski (approved):
lgtm

PR #80, opened 2021-03-12
cancels all orders set by iceberg algo on life stop event

avsek477:
This removes all of the respective atomic orders set by iceberg algo when stopped by the user.

### DIFF
--- a/lib/iceberg/events/life_stop.js
+++ b/lib/iceberg/events/life_stop.js
@@ -10,16 +10,20 @@
  */
 const onLifeStop = async (instance = {}) => {
   const { state = {}, h = {} } = instance
-  const { timeout } = state
-  const { debouncedSubmitOrders, debug, updateState } = h
+  const { orders = {}, gid, timeout } = state
+  const { debouncedSubmitOrders, emit, debug, updateState } = h
 
   debouncedSubmitOrders.cancel()
+
+  debug('detected iceberg algo cancelation, stopping...')
 
   if (timeout !== null) {
     clearTimeout(timeout)
     await updateState(instance, { timeout: null })
     debug('cleared timeout')
   }
+
+  await emit('exec:order:cancel:all', gid, orders)
 }
 
 module.exports = onLifeStop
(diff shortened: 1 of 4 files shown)


lib/iceberg/events/life_stop.js
```diff
@@ -9,10 +9,16 @@
...
+  const { state = {}, h = {} } = instance
+  const { args = {}, orders = {}, gid } = state
+  const { debouncedSubmitOrders, emit, debug } = h
+  const { cancelDelay } = args
 
   debouncedSubmitOrders.cancel()
+
+  debug('detected iceberg algo cancelation, stopping...')
+
+  await emit('exec:order:cancel:all', gid, orders, cancelDelay)
```
robertkowalski, line 21: should happen immediately, not with delay?

robertkowalski: set to `0`?

robertkowalski: hmmm, its a "feature" right now... no idea what the sense of it is

robertkowalski: `onLifeStop ` should have a cancel delay of 0, when we want to keep the feature, we need the separate cancel in `orders_order_cancel.js`

lib/iceberg/events/life_stop.js
```diff
@@ -10,16 +10,14 @@
...
-  const { debouncedSubmitOrders, debug, updateState } = h
+  const { orders = {}, gid } = state
+  const { debouncedSubmitOrders, emit, debug } = h
 
   debouncedSubmitOrders.cancel()
 
-  if (timeout !== null) {
-    clearTimeout(timeout)
-    await updateState(instance, { timeout: null })
-    debug('cleared timeout')
```
robertkowalski, line 21: you still have to clear the timeout from submit delays?

avsek477: thanks!! i will add this

robertkowalski (approved):
lgtm

PR #81, opened 2021-03-12
order not found issue fix

avsek477:
This fixes the error (`Order not found`) thrown when the algo server tries to cancel the already cancelled orders. This case occurs when a user tries to cancel an atomic order. The algo server when processing `order.close` event tries to cancel all the orders stored in its state based on the `id` of the order and not the `status` .


robertkowalski (approved):
lgtm

PR #83, opened 2021-03-16
Add safety net twap not skip current tick

avsek477:
The book updates asynchronously using an event emitter. So, this fix waits for the book to be updated in local after the order cancellation and then calculates the price with the new updated book without skipping the current tick.

### DIFF
--- a/lib/twap/events/self_interval_tick.js
+++ b/lib/twap/events/self_interval_tick.js
@@ -99,6 +99,7 @@ const onSelfIntervalTick = async (instance = {}) => {
 
   const order = generateOrder(state, orderPrice)
   if (order) {
+    await updateState(instance, { isBookUpdated: false })
     await emit('exec:order:submit:all', gid, [order], 0)
   }
 
--- a/lib/twap/events/self_interval_tick.js
+++ b/lib/twap/events/self_interval_tick.js
@@ -9,6 +9,7 @@ const generateOrder = require('../util/generate_order')
 const getOBPrice = require('../util/get_ob_price')
 const getTradePrice = require('../util/get_trade_price')
 const isTargetMet = require('../util/is_target_met')
+const scheduleTick = require('../util/schedule_tick')
 
 /**
  * Submits the next slice order if the price condition/target is met.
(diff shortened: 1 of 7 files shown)


lib/twap/events/self_interval_tick.js
```diff
@@ -71,11 +79,11 @@ const onSelfIntervalTick = async (instance = {}) => {
 
   let orderPrice
 
-  if (!/MARKET/.test(orderType)) {
+  if (!isMarketOrder) {
     if (hasTradeTarget(args)) {
       orderPrice = getTradePrice(state)
     } else if (hasOBTarget(args)) {
-      orderPrice = getOBPrice(state)
+      orderPrice = getOBPrice(_get(instance, 'state', {}))
```
robertkowalski, line 86: why using lodash get here? wouldn't we have a serious problem anyway if state is not set at this point?

lib/twap/events/self_interval_tick.js
```diff
@@ -42,9 +44,15 @@ const onSelfIntervalTick = async (instance = {}) => {
...
 
+  if (!isMarketOrder && !isBookUpdated) {
+    scheduleTick()
+    return
+  }
+
   if (!tradeBeyondEnd && !_isEmpty(orders)) {
     const [, t] = timeout(cancelDelay)
     await t()
+    await updateState(instance, { isBookUpdated: false })
```
robertkowalski, line 55: this is on cancel... but after sending an order, shouldn't the state change to `{ isBookUpdated: false }` too?

lib/twap/events/self_interval_tick.js
avsek477: Since we updated the state with `isBookUpdated` to be false, the local `state` which we defined before updating the state still has `isBookUpdated: true`. So, reading the state here from the `instance` reference. Lodash get was just a preference over `instance.state`. `getOBPrice` will return null if the book isn't updated resulting in orderPrice to be null and scheduleTick function will be called.

lib/twap/events/self_interval_tick.js
avsek477: will add this!!

lib/twap/events/self_interval_tick.js
```diff
@@ -69,18 +69,22 @@ const onSelfIntervalTick = async (instance = {}) => {
     await t()
   }
 
+  if (instance.state.shutdown) {
```
robertkowalski, line 72: 👍

lib/twap/events/self_interval_tick.js
```diff
@@ -102,10 +106,11 @@ const onSelfIntervalTick = async (instance = {}) => {
 
   const order = generateOrder(state, orderPrice)
   if (order) {
+    await updateState(instance, { isBookUpdated: false })
     await emit('exec:order:submit:all', gid, [order], 0)
```
robertkowalski, line 110: before this line a shutdown check would be great, too!

avsek477:
Solved in PR: https://github.com/bitfinexcom/bfx-hf-algo/pull/175

PR #85, opened 2021-03-17
start of algo orders: fix issues

robertkowalski:
in case the websocket is slow, and algo orders have not returned
a snapshot via the event emitter in the constructor yet, don't
crash the server

related code:
https://github.com/bitfinexcom/bfx-hf-algo/blob/ad890273da217362483c84505394af52b4d6ada7/lib/ao_host.js#L113-L114

### DIFF
--- a/lib/ao_host.js
+++ b/lib/ao_host.js
@@ -110,8 +110,11 @@ class AOHost extends AsyncEventEmitter {
 
     bindWS2Bus(this)
 
-    this.adapter.once('order:snapshot', (snapshot) => {
+    this.adapter.on('order:snapshot', (snapshot) => {
       this.orderSnapshot = snapshot
+    })
+
+    this.adapter.once('order:snapshot', (snapshot) => {
       this.emit('ready')
     })
   }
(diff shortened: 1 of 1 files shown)


lib/ao_host.js
```diff
@@ -508,6 +495,32 @@ class AOHost extends AsyncEventEmitter {
     return gid
   }
 
+  async maybeCancelExistingOrders (state, gid) {
+    const snapshot = this.orderSnapshot
+
+    if (!snapshot) return
+
+    if (Array.isArray(snapshot)) {
```
avsek477, line 503: this wont work!! the order snapshot is never an array but the instance of Order model always .. its an object that has `length` key when more than one atomic order is active and doesn't have the `length` key when only one order is active. Hence the reason why it was doubling ping pong orders for `orderAmount:1`. Also, the reason we are iterating using `i` to cancel each order is because the order snapshot is an instance of Order model but has keys as 0,1,2,3 depending on the number of active orders(i.e. when there are more than one atomic orders active). However, its a normal order object with gid when there is only one active atomic order. When there are none active orders, the snapshot object received does not have `gid` or `length` key.

robertkowalski:
hm, invalid, took another path

PR #86, opened 2021-03-18
fix algo order startup

robertkowalski:
this PR contains two fixes:

- fix: update order snapshots not just once, small fix for the event emitter

- bootstrap: error when no snapshot is received 

when there are connection issues, no snapshot is received. in
those cases we want to refuse the start of the algo order with an
error, because we don't know what state our account is in.

this error can be handled by the spawning server, but should be
an exception. the server should run `connect` on the host and
then wait for the ready event to get emitted, before starting the
first algo order.

### DIFF
--- a/lib/ao_host.js
+++ b/lib/ao_host.js
@@ -110,8 +110,11 @@ class AOHost extends AsyncEventEmitter {
 
     bindWS2Bus(this)
 
-    this.adapter.once('order:snapshot', (snapshot) => {
+    this.adapter.on('order:snapshot', (snapshot) => {
       this.orderSnapshot = snapshot
+    })
+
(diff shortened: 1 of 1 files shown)


lib/ao_host.js
```diff
@@ -110,8 +110,11 @@ class AOHost extends AsyncEventEmitter {
 
     bindWS2Bus(this)
 
-    this.adapter.once('order:snapshot', (snapshot) => {
+    this.adapter.on('order:snapshot', (snapshot) => {
       this.orderSnapshot = snapshot
+    })
+
+    this.adapter.once('order:snapshot', (snapshot) => {
```
avsek477, line 119: unused variable

avsek477 (approved):
lgtm :+1:

PR #87, opened 2021-03-18
check for the latest shutdown state in twap

avsek477:
check for the latest shutdown state after waiting for the submitDelay period preventing from submitting orders to be submitted if user cancelled algo order in between that time


robertkowalski (approved):
lgtm!

PR #88, opened 2021-03-19
await delay before submitting orders before generating orders

avsek477:
Awaits for the submit delay before generating new orders and read the amounts from the latest state if the adjacent order was filled in between the submit delay await. If the submit delay is relatively a substantial number, this prevents from submitting two new orders and gets the job done in one if the adjacent order was filled in between this submit delay awaiting period.

### DIFF
--- a/lib/iceberg/events/self_submit_orders.js
+++ b/lib/iceberg/events/self_submit_orders.js
@@ -15,12 +15,13 @@ const generateOrders = require('../util/generate_orders')
  */
 const onSelfSubmitOrders = async (instance = {}) => {
   const { state = {}, h = {} } = instance
-  const { emit, timeout } = h
+  const { emit, timeout, updateState } = h
   const { args = {}, gid } = state
   const { submitDelay } = args
 
   if (submitDelay) {
-    const [, t] = timeout(submitDelay)
+    const [id, t] = timeout(submitDelay)
+    await updateState(instance, { timeout: id })
     await t()
   }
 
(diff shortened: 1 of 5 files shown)


lib/iceberg/events/self_submit_orders.js
```diff
@@ -15,12 +15,18 @@ const generateOrders = require('../util/generate_orders')
...
   const { state = {}, h = {} } = instance
-  const { emit } = h
+  const { emit, timeout } = h
   const { args = {}, gid } = state
   const { submitDelay } = args
-  const orders = generateOrders(state)
 
-  return emit('exec:order:submit:all', gid, orders, submitDelay)
+  if (submitDelay) {
+    const [, t] = timeout(submitDelay)
```
robertkowalski, line 23: timeout needs to get stopped when stopping the order?

avsek477: thanks!! will add this

lib/iceberg/events/self_submit_orders.js
```diff
@@ -15,12 +15,19 @@ const generateOrders = require('../util/generate_orders')
...
   const { state = {}, h = {} } = instance
-  const { emit } = h
+  const { emit, timeout, updateState } = h
   const { args = {}, gid } = state
   const { submitDelay } = args
-  const orders = generateOrders(state)
 
-  return emit('exec:order:submit:all', gid, orders, submitDelay)
+  if (submitDelay) {
+    const [id, t] = timeout(submitDelay)
```
robertkowalski, line 23: timeout needs to get cleared on `life_stop.js`?

robertkowalski: something alike

https://github.com/bitfinexcom/bfx-hf-algo/blob/55db8d21ec2e6e4868e7c444a54a8228e7e809bd/lib/twap/events/life_stop.js#L12-L20

?

robertkowalski (changes requested):
found an issue when cancelling the algo and a submit delay was active

robertkowalski (approved):
lgtm

PR #89, opened 2021-03-22
Remove cancel delay

avsek477:
Remove the setTimeout function because no delay is required; and submit the orders for cancellation instantly.


robertkowalski (approved):
lgtm

PR #90, opened 2021-03-22
Remove cancel delays from ping pong orders

avsek477:
Remove the cancel delay options from ping pong orders.


robertkowalski (approved):
lgtm

PR #91, opened 2021-03-23
Remove cancel delay options from twap

avsek477:
- remove cancel delay options from twap orders
- handle cancel all orders on `stop` event
- remove cancelling all orders from `orders_cancel` event
- removed tests for cancelled delays


robertkowalski (approved):
lgtm

PR #93, opened 2021-03-23
Remove cancel delay acc dis

avsek477:
- remove cancel delay options from accumulate/distribute orders
- removed tests for cancelled delays


robertkowalski (approved):
lgtm

PR #96, opened 2021-03-23
ma crossover: fix cid, use cid helper

robertkowalski:
nonce generation was creating a nonce far too big:

https://github.com/bitfinexcom/bfx-api-node-util/blob/c7fab825b0f4bdf306a5286acd566a5a0936eb0c/lib/nonce.js#L6

error:

```
  bfx:hf:algo:bfx-ma_crossover:1616505422483 cid: invalid +43ms
```


avsek477 (approved):
lgtm

PR #97, opened 2021-03-25
handle "action: disabled" api response

robertkowalski:
in case there is an api error returned with "action disabled"
the algo order would keep running, no atomic orders would get
submitted and no error would be shown.

this PR handles `action: disabled` error responses and sends an
error message to the ui. it also stops the algo order that errors


robertkowalski:
superceded by #98

PR #98, opened 2021-03-25
ws-api: handle unknown errors

robertkowalski:
this catches all unknown errors returned by the api, so the algo
order is shut down properly and does not remain in idle state.

in case there is an api error returned with "action disabled"
the algo order would keep running, no atomic orders would get
submitted and no error would be shown.

this PR handles unknown error responses and sends an
error message to the ui. it also stops the algo order that errors

supercedes #97


avsek477 (approved):
<del> lgtm 
missed this before : https://github.com/bitfinexcom/bfx-hf-algo/pull/98#discussion_r602034379

lib/host/gen_helpers.js
```diff
@@ -231,6 +231,8 @@ const genHelpers = (state = {}, adapter) => {
 
             await adapter.submitOrderWithDelay(connection, retryDelay, order)
           }
+        } else {
+          await state.ev.emit('error:unknown', gid, order, { text })
```
avsek477, line 235: should be `error:unknown_error`

PR #100, opened 2021-03-26
Submit active orders for cancellations

avsek477:
If the status isn't `ACTIVE` means that the order wasn't submitted successfully in the first place and is not required to be cancelled.

### DIFF
--- a/lib/host/events/cancel_all_orders.js
+++ b/lib/host/events/cancel_all_orders.js
@@ -2,6 +2,7 @@
 
 const _isObject = require('lodash/isObject')
 const withAOUpdate = require('../with_ao_update')
+const excludeOrderStatus = ['CANCELED', 'EXECUTED']
 
 /**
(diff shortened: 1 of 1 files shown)


lib/host/events/cancel_all_orders.js
```diff
@@ -22,7 +22,7 @@ module.exports = async (aoHost, gid, orders) => {
 
     // Don't try to cancel market orders
     const _orders = allOrders
-      .filter(o => !/MARKET/.test(o.type) && o.id && !/CANCELED/.test(o.status))
+      .filter(o => !/MARKET/.test(o.type) && o.id && /ACTIVE/.test(o.status))
```
robertkowalski, line 25: sadly that does not work, here is the status of a partially filled order:

```
  'PARTIALLY FILLED @ 36764.0(-0.01)',
```

vigan-abd (approved):

PR #102, opened 2021-03-29
Delete p iteration

vitaliystoliarovcc:
Asana ticket: https://app.asana.com/0/1125859137800433/1200116176160281/f


lib/async_event_emitter.js
```diff
@@ -124,18 +123,20 @@ class AsyncEventEmitter {
   async emit (eventName, ...args) {
     const indexesToRemove = []
 
-    await PI.forEachSeries(this.listeners, async (l, i) => {
+    for (const i in this.listeners) {
```
avsek477, line 126: better to use for .. of loop here

avsek477 (changes requested):
there's usage of p-iteration in tests... it's best to remove it's usage from tests as well.. the tests are currently successful because some dependencies in this repo installs the module for us

lgtm other than this.. nice work :)

lib/async_event_emitter.js
```diff
@@ -124,18 +123,20 @@ class AsyncEventEmitter {
...
 
-        if (l.type === 'once') {
+        if (listener.type === 'once') {
           indexesToRemove.push(i)
         }
       }
-    })
+    }
 
     indexesToRemove.forEach(i => this.listeners.splice(i, 1))
```
avsek477, line 141: I think this is a bug.. would have to check properly though.. Might not work if `indexesToRemove` has more than one indices to splice.. since splice replaces the previous indexes

avsek477, line 141: I think this is a bug.. would have to check properly though.. Might not work if indexesToRemove has more than one indices to splice.. since splice replaces the previous indexes

vitaliystoliarovcc: Yes, you are right, the problem may appear. 
I think the solution would be filtering and redefining this.listeners or sorting of indexesToRemove  in descending order

robertkowalski:
testsuite does not pass any more:

```js
// /test/lib/accumulate_distribute/meta/declare_channels.js:

const assert = require('assert')
const PI = require('p-iteration')
const declareChannels = require('../../../../lib/accumulate_distribute/meta/declare_channels')

```

robertkowalski (changes requested):
testsuite should pass

lib/async_event_emitter.js
avsek477: better to avoid `for .. in` loop as well then for arrays(not really a fan of using them in arrays anyway) .. will have to research a bit but using `for .. in` loop for arrays isn't recommended

vitaliystoliarovcc:
Wow.. didn't notice it since the dependency `bfx-hf-ext-plugin-bitfinex` has `p-iteration` as a dependency and tests that use it pass successfully
I'll replace forEachSeries in the tests as well

lib/ao_host.js
```diff
@@ -753,11 +752,11 @@ class AOHost extends AsyncEventEmitter {
...
    */
   async triggerGlobalEvent (section, eventName, ...args) {
-    const instances = Object.values(this.instances)
+    const instances = this.getAOInstances()
 
-    return PI.forEach(instances, async (instance) => (
-      this.triggerAOEvent(instance, section, eventName, ...args)
-    ))
+    await Promise.all(instances.map(async instance => {
+      await this.triggerAOEvent(instance, section, eventName, ...args)
```
robertkowalski, line 758: this one is tricky, some code sections use `await triggerGlobalEvent` and would be delayed by it now

robertkowalski: https://github.com/bitfinexcom/bfx-hf-algo/blob/4220bed6f6cbb87655fb38781ad743db4a790082/lib/host/ws2/process_message.js#L130-L144

robertkowalski (changes requested):
to keep current behaviour probably the promise.all call just commented is not required, to keep the current behaviour a simple Array.prototype.forEach can be used

lib/ao_host.js
vitaliystoliarovcc: > would be delayed by it now

so shouldn't it "await" for `triggerAOEvent` as it was done with PI.forEach before? This can significantly change the behavior of the system, since in the master branch it waits for all events to be sent

robertkowalski: original code did not await, thats the issue here

```js
    return PI.forEach(instances, async (instance) => (
      this.triggerAOEvent(instance, section, eventName, ...args)
    ))
```

vitaliystoliarovcc:
```js
async (instance) => ( this.triggerAOEvent(instance, section, eventName, ...args) ))
```

is equal to 
```js
async (instance) => {
  return this.triggerAOEvent(instance, section, eventName, ...args)
 }
```

robertkowalski: oh its using `()` -- my bad, i missed that

robertkowalski (approved):
lgtm

PR #103, opened 2021-03-30
fix memleak on closing algo orders

robertkowalski:
when an algo order was stopped, events were still listened for,
triggering event updates.

to reproduce:
1. start a twap, delay 0s
2. stop it
3. start another one, delay 0s
4. stop it
5. start another one set delay to 5s

you have 3 save events for 3 algo orders in the log when order
book updates arrive for the last algo order.


avsek477 (approved):
lgtm

PR #104, opened 2021-03-30
twap: fix delay on order submit

robertkowalski:
chained events behave synchronous, so we can't "wait" for the
first interval to "finish"


avsek477 (approved):
lgtm

PR #105, opened 2021-03-30
iceberg: fix delay on order submit

avsek477:
do not wait for the first interval to finish on order submit


robertkowalski (approved):
lgtm

PR #106, opened 2021-03-31
allow cancelling orders before removing all listeners

avsek477:
- remove remaining cancel order listeners on algo stop
- fix for not removing all listeners before cancelling existing orders set by algo


robertkowalski (approved):
lgtm

PR #108, opened 2021-04-02
Fix subscription dups

vitaliystoliarovcc:
Asana ticket: https://app.asana.com/0/1125859137800433/1200136786490173/f

Added a check to prevent duplicate subscription with the same parameters


avsek477 (approved):
lgtm :+1:

PR #110, opened 2021-04-05
Display price only if orderType is LIMIT

vitaliystoliarovcc:
Asana ticket: https://app.asana.com/0/1125859137800433/1200130365695551/f

The issue is related to the construction of the order label. This is purely a view problem

The price was always displayed after '@' when it was greater than 0
Now it will be displayed only if _Order Type_ is LIMIT


lib/ma_crossover/meta/gen_order_label.js
```diff
@@ -13,10 +13,11 @@
 const genOrderLabel = (state = {}) => {
   const { args = {} } = state
   const { orderType, orderPrice, amount, long, short } = args
+  const priceShouldBeDisplayed = orderType === 'LIMIT'
 
   return [
     'MA Crossover',
-    ` | ${amount} @ ${orderPrice || orderType} `,
+    ` | ${amount} @ ${priceShouldBeDisplayed ? orderPrice : orderType} `,
```
avsek477, line 20: May be we should add the following in `process_params.js` as well so that we dont save it in db. What do you think @vitaliystoliarovcc ?
```
if (params.orderType !== 'LIMIT') {
    params.orderPrice = 0
}
```

vitaliystoliarovcc: I don’t think this is necessary, as it is assumed that orderPrice in the Birfinex API will be ignored if orderType !== 'LIMIT'. @robertkowalski, isn't?

vitaliystoliarovcc: > as well so that we don't save it in db

can you point where this data is stored in the database?

robertkowalski: > can you point where this data is stored in the database?

```
~/.honeyframework/algos.json
```

is a json database of algo orders. it contains the data serialized in `meta/serialize.js`

robertkowalski: > I don’t think this is necessary, as it is assumed that orderPrice in the Birfinex API will be ignored if orderType !== 'LIMIT'. @robertkowalski, isn't?

hmmm... in the first place i was wondering why the ui allows to submit market orders with a price...? then as second thought i think the cleaner approach is to fix it in process params, if we receive invalid data from the ui. for the feature the current approach also makes things working but why should we keep invalid values?

robertkowalski (approved):
lgtm

PR #111, opened 2021-04-06
Fix async event emitter

vitaliystoliarovcc:
⚠️ This PR is based on changes in https://github.com/bitfinexcom/bfx-hf-algo/pull/102

Fixes:
- fix removing of unwanted listeners ([Asana ticket](https://app.asana.com/0/1125859137800433/1200129837232907/f))
- fix emit of one time listeners on race condition (can happen when emit is called at the same time for a specific one-time listener)

### DIFF
--- a/lib/ao_host.js
+++ b/lib/ao_host.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const PI = require('p-iteration')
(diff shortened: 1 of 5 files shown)


avsek477 (approved):
lgtm

robertkowalski (approved):
lgtm

PR #95, opened 2021-03-23
Remove cancel delay ococo

avsek477:
- remove cancel delay options from oco algo
- removed tests for cancelled delays

Related PR: https://github.com/bitfinexcom/bfx-hf-algo/pull/89

PR #94, opened 2021-03-23
Remove cancel delay ma crossover

avsek477:
- remove cancel delay options from ma crossover orders
- removed tests for cancelled delays

Related PR: https://github.com/bitfinexcom/bfx-hf-algo/pull/89

PR #112, opened 2021-04-08
Fix waiting for order submitting

vitaliystoliarovcc:
Asana tickets:

- https://app.asana.com/0/1125859137800433/1200162041515069/1200163628599973/f
- https://app.asana.com/0/1125859137800433/1200174906704562/f

Depends on: https://github.com/bitfinexcom/bfx-hf-algo/pull/121

The issue occurred when submitting orders in interval when the Internet connection is unstable or lost

Now, in case of an unstable or lost connection, the next tick of the interval will not be emited until async submit function is executed

---

In addition, in the method for orders bulk submit, the sequential sending of requests was replaced with parallel. The result of the implementation  is shown in the attached video

https://user-images.githubusercontent.com/70510980/115058628-2976f900-9eee-11eb-85e2-8dce99686da8.mov

### DIFF
--- a/lib/host/gen_helpers.js
+++ b/lib/host/gen_helpers.js
@@ -54,8 +54,8 @@ const genHelpers = (state = {}, adapter) => {
    * * `cancelOrder(state, order)` - takes current algo state and cancels the
    *   provided order
    *
-   * * `submitOrderWithDelay(state, delay, order)` - takes current algo state,
-   *   submits a new order, delay in ms
+   * * `submitOrder(state, order)` - takes current algo state,
+   *   submits a new order
    *
    * * `declareEvent(instance, host, eventName, path)` - declares an internal
    *   AO event, see section below
@@ -125,7 +125,7 @@ const genHelpers = (state = {}, adapter) => {
      * @param {...any} eventArgs - args passed to all handlers
      * @returns {Promise} p - resolves when all handlers complete
      * @example
-     * await emit('exec:order:submit:all', gid, [order], submitDelay)
+     * await emit('exec:order:submit:all', gid, [order])
      */
     emit: async (eventName, ...eventArgs) => {
       logHelperCall('emit %s', eventName)
@@ -189,22 +189,21 @@ const genHelpers = (state = {}, adapter) => {
     },
 
     /**
-     * Submits an order after a delay, and adds it to the active order set on
+     * Submits an order and adds it to the active order set on
      * the AO state. Emits errors if the order fails to submit; retries up to
      * MAX_SUBMIT_ATTEMPTS in the case of balance evaluation errors.
      *
      * @param {object} state - current AO instance state
-     * @param {number} delay - delay in milliseconds
      * @param {object} order - order model
      * @param {number} attempt - attempt count, max is 10 (set as a constant)
      * @returns {object} nextState
      * @example
-     * await submitOrderWithDelay(state, 100, order)
+     * await submitOrder(state, order)
      */
-    submitOrderWithDelay: async (state = {}, delay, order, attempt = 1) => {
+    submitOrder: async (state = {}, order, attempt = 1) => {
       logHelperCall(
-        'submitOrderWithDelay (%dms, attempt %d): %s',
-        delay, attempt, order.toString()
+        'submitOrder (attempt %d): %s',
+        attempt, order.toString()
       )
 
       if (attempt > MAX_SUBMIT_ATTEMPTS) { // note state assumed already patched
@@ -216,10 +215,9 @@ const genHelpers = (state = {}, adapter) => {
         [order.cid + '']: order
       }
 
-      // Note that we don't wait for the order to submit here, since it might
-      // fill immediately triggering a fill event before we patch the state
-      // orders/allOrders objects
-      await adapter.submitOrderWithDelay(connection, delay, order).then((order) => { // eslint-disable-line
+      // FIXME
+      // add "await" for the function to track request to the Bitfinex API
+      adapter.submitOrder(connection, order).then((order) => { // eslint-disable-line
         debug(
           'order successfully submitted: %s %f @ %f %s',
           order.type, order.amountOrig, order.price, order.status
@@ -237,9 +235,7 @@ const genHelpers = (state = {}, adapter) => {
           if (attempt === MAX_SUBMIT_ATTEMPTS) {
             await state.ev.emit('error:evaluate_balance', gid, order, { text })
           } else {
-            const retryDelay = Math.max(500, delay + 250) // min 0.5s, increasing
-
-            await adapter.submitOrderWithDelay(connection, retryDelay, order)
+            await adapter.submitOrder(connection, order)
           }
         } else {
           await state.ev.emit('error:unknown_error', gid, order, { text })
--- a/lib/host/gen_helpers.js
+++ b/lib/host/gen_helpers.js
@@ -146,6 +146,17 @@ const genHelpers = (state = {}, adapter) => {
       await state.ev.emit('notify', gid, level, message)
     },
 
+    subscribeDataChannels: (state = {}) => {
+      const { channels = [], gid, connection } = state
+
+      channels.forEach((ch) => {
+        debug('subscribing to channel %j [AO gid %d]', ch, gid)
+
+        // TODO: optional await
+        adapter.subscribe(connection, ch.channel, ch.filter)
+      })
+    },
+
     /**
      * Cancels the provided order instantly, and removes it from the active
      * order set.
@@ -186,6 +197,17 @@ const genHelpers = (state = {}, adapter) => {
           [order.cid + '']: order
         }
       }
+    },
+
+    sendPing: async (state = {}) => {
+      const { connection } = state
+
+      logHelperCall(
+        'sending ping to get server time'
+      )
+
+      const data = await adapter.sendPing(connection)
+      return data
     },
 
     /**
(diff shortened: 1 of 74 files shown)


lib/host/gen_helpers.js
```diff
@@ -219,7 +219,7 @@ const genHelpers = (state = {}, adapter) => {
       // Note that we don't wait for the order to submit here, since it might
       // fill immediately triggering a fill event before we patch the state
       // orders/allOrders objects
-      adapter.submitOrderWithDelay(connection, delay, order).then((order) => { // eslint-disable-line
+      await adapter.submitOrderWithDelay(connection, delay, order).then((order) => { // eslint-disable-line
```
avsek477, line 222: we can't await here.. because as a result of order submission,  we receive different events like `order_new`, `order_fill` and `order_cancel`  which is being handled here `https://github.com/bitfinexcom/bfx-hf-algo/blob/master/lib/host/ws2/process_message.js` to update the order statuses in state which is required for algo orders to process them accordingly

vitaliystoliarovcc: @avsek477  this is necessary to make sure that the request has been performed, otherwise it will always be considered completed even if there is no Internet connection

robertkowalski: hm, that would block the whole execution path, indeed we can't wait here...

people can currently define a "submit delay". its measured in seconds. when you define a submit delay of 30 seconds the promise won't resolve for that timeframe

robertkowalski: i agree with @vitaliystoliarovcc, too

where we need parallelism in order handling, we could use things like async.parallel / promise.all and keep a map of current open orders. the current sequential emulation we have is too limited, but also has its uses for simple algos

robertkowalski: suggestion: we remove the "feature" submitDelay - similar like we did remove cancel-delays in the last release. the feature has no practical use and complicates the code base. after that is done, we add concurrent handling to the algo orders that need it. for example twap orders, there is just one open order, so we can go with "awaiting" a successful order submit.

ping/pong orders support multiple order submissions at a time. for that the function should take an array of orders and resolve only when all orders are submitted (in case some are cancelled e.g. because of invalid balance it should also fullfill the promise, am i am thinking correct here?)

so two separate tasks:

1. remove submitDelay in all algo orders, submit one PR per algo order, we can hardcode submit delay to 0 in this intermediate step
2. change the submit order function to: not having any delay functionality, accept multiple orders for bulk submits

this way we solve a few issues at once we have with order submits and can also resolve the "no-wifi" issue. does the plan sound good to you @avsek477 @vitaliystoliarovcc ?

vitaliystoliarovcc: > similar like we did remove cancel-delays

can you share the link to the PR?

robertkowalski: https://github.com/bitfinexcom/bfx-hf-algo/pull/90
https://github.com/bitfinexcom/bfx-hf-algo/pull/91
https://github.com/bitfinexcom/bfx-hf-algo/pull/92
https://github.com/bitfinexcom/bfx-hf-algo/pull/93
https://github.com/bitfinexcom/bfx-hf-algo/pull/94
https://github.com/bitfinexcom/bfx-hf-algo/pull/95

vitaliystoliarovcc: > (in case some are cancelled e.g. because of invalid balance it should also fullfill the promise, am i am thinking correct here?)

Makes sense to me


As I understood, the main problem with "await" is the inclusion of a delay in the submitOrderWithDelay helper, so using a loop, each next order will wait for all previous ones.

For example, the app emits
```
emit('exec:order:submit:all', gid, [A,B,C], submitDelay)
```
where A, B, C - orders

1. Without _await_ (changes in `master` branch - NOT OK, the submit responses are ignored)
```
A->++++++>==>
   B->++++++>=========>X - Ooops, request timeout
      C->++++++>==> 

/// Interval's promise:
--------> // Promise resolved despite the request is still running
```

2. With _await_ (this PR's changes - NOT OK, each next order in list will wait for the delay of the previous one)

```
A->++++++>==>
             B->++++++>=========>X - Ooops, request timeout
                                 C->++++++>==> 

/// Interval's promise:
---------------------------------------------> // Promise resolved when all request are fulfilled, but it takes to long if we submit more than one order
```

3. With _await_ and Promise.all (TODO)

```
A->++++++>==>
B->++++++>=========>X - Ooops, request timeout
C->++++++>==> 

/// Interval's promise:
----------------------> // Promise resolved in time equal to the longest order submission
```

Legend: 

-> - Promise resolving
++++> = Delay
==> Request

@robertkowalski @avsek477 is it correct?

The simplest solution I see is to replace a loop with Promise.all in [submit_all_orders.js](https://github.com/bitfinexcom/bfx-hf-algo/blob/master/lib/host/events/submit_all_orders.js)

What do you think?

vitaliystoliarovcc: > we remove the "feature" submitDelay
> remove submitDelay in all algo orders, submit one PR per algo order, we can hardcode submit delay to 0 in this intermediate step

But if we remove `submitDelay`, it will affect some orders:
MA Crossover, OcO, Ping/Pong and Accumulate/Distribute

"Delay" can be moved next to `emit('exec:order:submit:all')`, as it was done for Iceberg

https://github.com/bitfinexcom/bfx-hf-algo/blob/405abcbb8b10c0686246ab707590c8834bad0266/lib/iceberg/events/self_submit_orders.js#L22-L29

But will this work for all types of Algo orders? Perhaps it will be important to perform some validation (inside `submitOrderWithDelay` in bfx-hf-algo/lib/host/gen_helpers.js and bfx-hf-algo/lib/ws_adapter.js) before delay

robertkowalski: > With await and Promise.all (TODO)
> @robertkowalski @avsek477 is it correct?

yes this is correct

> But will this work for all types of Algo orders? Perhaps it will be important to perform some validation (inside submitOrderWithDelay in bfx-hf-algo/lib/host/gen_helpers.js and bfx-hf-algo/lib/ws_adapter.js) before delay

all algo orders will need to get touched and then tested. best to start with TWAP for now and submit a PR for it. algo orders behaviour need to be studied carefully, they can be tricky sometimes. ping/pong might need more in-depth changes to the acutal code, when you try algo orders currently you can select an order count > 1, in those cases many orders will be sent to the server concurrently and stay open in parallel. so each algo must be tested carefully before making the changes

vitaliystoliarovcc: > best to start with TWAP for now and submit a PR for it.

if I make changes in submit_all_orders.js, all order types will be affected

avsek477: > > best to start with TWAP for now and submit a PR for it.
> 
> if I make changes in submit_all_orders.js, all order types will be affected

We can create a new event similar to `exec:order:submit:all` in `ao_host.js` and use the new event for submission of orders for twap orders for now

vitaliystoliarovcc:
According to https://github.com/bitfinexcom/bfx-hf-algo/pull/121#discussion_r614220863, Promise.all should be implemented in such a way as to to prevent re-occurrence of the issue https://github.com/bitfinexcom/bfx-hf-algo/pull/121#discussion_r613888851 when "await" is present

vitaliystoliarovcc:
Current issue: https://app.asana.com/app/asana/-/get_asset?asset_id=1200206056851202

vitaliystoliarovcc:
image
image

avsek477 (approved):
lgtm

lib/host/gen_helpers.js
```diff
@@ -259,28 +253,12 @@ const genHelpers = (state = {}, adapter) => {
           if (attempt === MAX_SUBMIT_ATTEMPTS) {
             await state.ev.emit('error:evaluate_balance', gid, order, { text })
           } else {
-            const retryDelay = Math.max(500, delay + 250) // min 0.5s, increasing
-
-            await adapter.submitOrderWithDelay(connection, retryDelay, order)
+            await adapter.submitOrder(connection, order)
```
robertkowalski, line 256: should we wait a bit here? how about 500ms?

lib/host/gen_helpers.js
```diff
@@ -211,37 +211,31 @@ const genHelpers = (state = {}, adapter) => {
...
 
       const { connection } = state
-      const orderPatch = {
-        [order.cid + '']: order
-      }
 
-      // Note that we don't wait for the order to submit here, since it might
-      // fill immediately triggering a fill event before we patch the state
-      // orders/allOrders objects
-      adapter.submitOrderWithDelay(connection, delay, order).then((order) => { // eslint-disable-line
```
robertkowalski, line 244: did you check this note?

```
      // Note that we don't wait for the order to submit here, since it might
      // fill immediately triggering a fill event before we patch the state
      // orders/allOrders objects
```

lib/host/gen_helpers.js
vitaliystoliarovcc: this part was fixed in https://github.com/bitfinexcom/bfx-hf-algo/pull/121/commits/2543ced887c50a6734d355c99041461b48807e8c

lib/host/gen_helpers.js
robertkowalski: just checking, this is related to this LOC in submit_all_orders.js:

```js
nextState = await submitOrderWithDelay(nextState, delay, o)
```

robertkowalski: covered by the internal `withAOUpdate` now, so we are good

vitaliystoliarovcc: Of course. This is no longer required since the loop has been replaced with Promise.all 

https://github.com/bitfinexcom/bfx-hf-algo/blob/a0b4bceacf9bbe34605e1c28e4098826b41e5008/lib/host/events/submit_all_orders.js#L51-L58

vitaliystoliarovcc:
Ahh... conflicts again. Fixing...

robertkowalski (approved):
lgtm

PR #113, opened 2021-04-09
Refactor accumulate distribute delay

vitaliystoliarovcc:
- fixed submitDelay format (from seconds to ms)
- perform 'timeout' before exec:order:submit:all

### DIFF
--- a/lib/accumulate_distribute/events/self_submit_order.js
+++ b/lib/accumulate_distribute/events/self_submit_order.js
@@ -17,15 +17,21 @@ const generateOrder = require('../util/generate_order')
  */
 const onSelfSubmitOrder = async (instance = {}) => {
   const { state = {}, h = {} } = instance
-  const { emit, debug } = h
+  const { emit, debug, timeout, updateState } = h
   const { args = {}, gid } = state
   const { submitDelay } = args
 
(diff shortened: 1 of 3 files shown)


lib/accumulate_distribute/events/self_submit_order.js
```diff
@@ -17,15 +17,21 @@ const generateOrder = require('../util/generate_order')
...
+  const { emit, debug, timeout, updateState } = h
   const { args = {}, gid } = state
   const { submitDelay } = args
 
   const order = generateOrder.gen(instance)
 
   if (order) {
+    if (submitDelay) {
+      const [id, t] = timeout(submitDelay)
+      await updateState(instance, { timeout: id })
```
avsek477, line 29: the timeout here replaces the one we are using for `self-interval` tick and the `self-interval` tick again replaces the timeout object set during `self-submit` resulting in memleak when we stop the algo order.. also we might wanna wait for the submit delay timeout before generating the order here since meanwhile we are awaiting for submit delay, the `order_fill` event might change some params on state

robertkowalski:
afaik we want to remove submit delay completely, so this refactor is not needed?

PR #117, opened 2021-04-12
Remove TWAP delay

vitaliystoliarovcc:
Asana ticket: https://app.asana.com/0/1125859137800433/1200174906704560/f

### DIFF
--- a/lib/twap/events/self_interval_tick.js
+++ b/lib/twap/events/self_interval_tick.js
@@ -24,7 +24,7 @@ const onSelfIntervalTick = async (instance = {}) => {
   const { orders = {}, args = {}, gid, shutdown } = state
   const { emit, debug, timeout, updateState, emitSelf } = h
   const {
-    priceTarget, tradeBeyondEnd, amount, sliceAmount, submitDelay, priceDelta,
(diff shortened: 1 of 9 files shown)


lib/twap/events/self_interval_tick.js
```diff
@@ -62,11 +62,6 @@ const onSelfIntervalTick = async (instance = {}) => {
     }
   }
 
-  if (submitDelay) {
-    const [, t] = timeout(submitDelay)
-    await t()
-  }
-
   if (instance.state.shutdown) {
```
avsek477, line 65: i think this check isn't necessary now since we aren't awaiting delay

vitaliystoliarovcc: It has async call `exec:order:cancel:all`
https://github.com/bitfinexcom/bfx-hf-algo/blob/8d0ec3f5e6af47beeab9a0bf03fcf451a4a36108/lib/twap/events/self_interval_tick.js#L41-L47

Perhaps the `state.shutdown` could be changed while the async functions are running before this check

robertkowalski (approved):
lgtm

PR #120, opened 2021-04-13
Validate params iceberg

avsek477:
This PR adds the functionality to be able to fetch the error fields as well so that it's easier in the UI side to show the error message on the specific field of the algo order. It can be used by UI to check for errors before submitting the order to algo server. However, the UI if wishes to use this function, it must process params first using `processParams` function before using the `validateParams` function. 

Related PR: https://github.com/bitfinexcom/bfx-hf-server/pull/96 to not break the existing function for params validation.

Task: https://app.asana.com/0/1125859137800433/1200038332394566

### DIFF
--- a/lib/host/init_ao_state.js
+++ b/lib/host/init_ao_state.js
@@ -2,6 +2,7 @@
 
 const clientID = require('../util/gen_client_id')
 const _isFunction = require('lodash/isFunction')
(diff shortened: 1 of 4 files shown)


vitaliystoliarovcc (approved):
lgtm

robertkowalski (approved):
lgtm

PR #121, opened 2021-04-13
Remove delay in submit order

vitaliystoliarovcc:


avsek477 (approved):
lgtm

lib/host/gen_helpers.js
```diff
@@ -216,10 +215,7 @@ const genHelpers = (state = {}, adapter) => {
         [order.cid + '']: order
       }
 
-      // Note that we don't wait for the order to submit here, since it might
-      // fill immediately triggering a fill event before we patch the state
-      // orders/allOrders objects
-      adapter.submitOrderWithDelay(connection, delay, order).then((order) => { // eslint-disable-line
+      await adapter.submitOrder(connection, order).then((order) => { // eslint-disable-line
```
robertkowalski, line 218: does this work with ping pong orders, submit delay 5seconds?

afaik this would be only possible with step 2 of the task, where we have concurrent order submits?

robertkowalski: example: order is submitted, confirmation from the backend takes 4 seconds to arrive...

robertkowalski: yes, confirmed... `submit_all_orders.js` calls `submitOrder` in a blocking loop

vitaliystoliarovcc: "await" has been removed to allow Ping/Pong order can create a large number of orders in parallel. 

 In this case, the problem [TWAP: Saving orders on state in each interval leads to numerous orders being submitted at once when internet connnection is restored.](https://app.asana.com/0/1125859137800433/1200162041515069/1200163628599973/f) remains and will be fixed in https://github.com/bitfinexcom/bfx-hf-algo/pull/112, but requires additional changes so as not to break the changes of current PR

lib/host/gen_helpers.js
```diff
@@ -237,9 +235,7 @@ const genHelpers = (state = {}, adapter) => {
           if (attempt === MAX_SUBMIT_ATTEMPTS) {
             await state.ev.emit('error:evaluate_balance', gid, order, { text })
           } else {
-            const retryDelay = Math.max(500, delay + 250) // min 0.5s, increasing
-
-            await adapter.submitOrderWithDelay(connection, retryDelay, order)
+            await adapter.submitOrder(connection, order)
```
robertkowalski, line 238: this needs some exponential backoff no?

vitaliystoliarovcc: why is it necessary?

robertkowalski: when the error `evaluate_balance` is received we need to wait a bit until the final balance is calculated.

without a backoff we the code just hammers the backend until `MAX_SUBMIT_ATTEMPTS` are reached wituot giving time to calculate the final balance

vitaliystoliarovcc: > to wait a bit until the final balance is calculated.

Ok, got it. Is the value from `Math.max(500, delay + 250)` suitable for delay? Perhaps it must be less than 500ms or more at first attempt?

vitaliystoliarovcc: Initially 'attempts` was implemented in https://github.com/bitfinexcom/bfx-hf-algo/commit/9d46d2be1aa7c802bfcc0c60c70fda247ce4aa68#

But did it really work? `submitOrderWithDelay` helper calls `submitOrderWithDelay` from `bfx-api-node-core` and passes incremented `attempt`, but [submit_with_delay.js](https://github.com/bitfinexcom/bfx-api-node-core/blob/6eac4278e3a01ca99dc9634987cf5a7f27e51b9e/lib/ws2/orders/submit_with_delay.js)
doesn't have any `attempt` parameter. 

Perhaps this was present in a particular version of the package `bfx-api-node-core`.
Even in this repo's `master` branch, the `adapter.submitOrderWithDelay` method can be called on fail, and `attempt` parameter will never be inremented. Can you confirm it?

Then I think it should be solved in another PR. In this one I will only return back the `retryDelay`

avsek477: yeah didnt notice this .. `attempt` doesn't work here.. 

another PR sounds good to me..

vitaliystoliarovcc: At the first attempt it [prints the Error](https://github.com/bitfinexcom/bfx-hf-algo/blob/01dd7332d28f02e948df72af083f5bb9a1a3392c/lib/host/gen_helpers.js#L196)
If the Error contains `evaluate`, it will call submitOrder second time and an Unhandled rejection will occur in case of error.
I'll create the ticket

robertkowalski (approved):
lgtm

PR #122, opened 2021-04-19
issue fix for the sliced order amounts not being distorted

avsek477:
This PR fixes the issue of not distorting some of the sliced order amounts. The order amount is always distorted and minimum value between the remaining and distorted amount is taken in the last iteration while generating order amounts. 

Task: https://app.asana.com/0/1125859137800433/1200207564508439

### DIFF
--- a/lib/accumulate_distribute/util/gen_order_amounts.js
+++ b/lib/accumulate_distribute/util/gen_order_amounts.js
@@ -26,7 +26,7 @@ const generateOrderAmounts = (args = {}) => {
 
     while (Math.abs(amount - totalAmount) > DUST) {
       const m = Math.random() > 0.5 ? 1 : -1
-      const orderAmount = Math.min(sliceAmount, sliceAmount * (1 + (Math.random() * amountDistortion * m)))
+      const orderAmount = sliceAmount * (1 + (Math.random() * amountDistortion * m))
       const remAmount = amount - totalAmount
       const cappedOrderAmount = +prepareAmount(remAmount < 0
         ? Math.max(remAmount, orderAmount)


lib/accumulate_distribute/util/gen_order_amounts.js
```diff
@@ -26,7 +26,7 @@ const generateOrderAmounts = (args = {}) => {
 
     while (Math.abs(amount - totalAmount) > DUST) {
       const m = Math.random() > 0.5 ? 1 : -1
-      const orderAmount = Math.min(sliceAmount, sliceAmount * (1 + (Math.random() * amountDistortion * m)))
```
vitaliystoliarovcc, line 29: One year ago `Math.min` was added in the commit https://github.com/bitfinexcom/bfx-hf-algo/commit/45663868971291b5f75ecb6cd0c7bd40aa4980d0#diff-56ca5cc772b02af17459607b243e9a3a9998659ba3f174b1e1daf03bd38d1a88R17

I assume it fixes the sum of all slices that could be a different value from the specified amount (

For example, without `Math.min` I received (only Sell):

image

The sum isn't equal to 60, but it should. @avsek477 @robertkowalski right?

avsek477: yes .. it should ... let me check again.. i tried with many example at last it would calculate whatever was remaining

vitaliystoliarovcc: One more example:
image

robertkowalski: after chat discussion: minimum order size amount will be solved in a separate task

robertkowalski (approved):
lgtm

PR #123, opened 2021-04-19
not accepting order amount if it isn't a number and equals 0

avsek477:
Checks for the value of order amount to be finite and not equals to 0 in ma crossover algo.

### DIFF
--- a/lib/ma_crossover/meta/validate_params.js
+++ b/lib/ma_crossover/meta/validate_params.js
@@ -41,7 +41,7 @@ const validateParams = (args = {}) => {
   } = args
 
   if (!_includes(ORDER_TYPES, orderType)) return `Invalid order type: ${orderType}`
-  if (!_isFinite(amount)) return 'Invalid amount'
+  if (!_isFinite(amount) || amount === 0) return 'Invalid amount'
   if (orderType === 'LIMIT' && !_isFinite(orderPrice)) {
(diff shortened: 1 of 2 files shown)


lib/ma_crossover/meta/validate_params.js
```diff
@@ -41,7 +41,7 @@ const validateParams = (args = {}) => {
   } = args
 
   if (!_includes(ORDER_TYPES, orderType)) return `Invalid order type: ${orderType}`
-  if (!_isFinite(amount)) return 'Invalid amount'
+  if (!_isFinite(amount) || amount === 0) return 'Invalid amount'
```
vitaliystoliarovcc, line 44: I think there should be a check: `amount > 0`

vitaliystoliarovcc (changes requested):

lib/ma_crossover/meta/validate_params.js
avsek477: so before we validate the params .. we are processing them first which negates the value if its a sell order .. so a check for amount equals 0 is fine here

vitaliystoliarovcc: I see. You are right

vitaliystoliarovcc (approved):
lgtm

robertkowalski (approved):
lgtm

PR #124, opened 2021-04-19
check for amount and sliceAmount values to be finite and nonzero

avsek477:
Checks for the value of order amount and sliced order amount to be finite and not equals to 0 in acc/dis algo.


vitaliystoliarovcc (approved):
lgtm

robertkowalski (approved):
lgtm

PR #125, opened 2021-04-19
ma-cross: guard against outdated data

robertkowalski:
the ws api sends a snapshot in the beginning. it contains historic
candles from the past. additionally, when there is a lot of traffic,
the ws2 sends the last 1-3 candles that are not part of the
snapshot yet.

example, on a 5min scale candle subscription:

1. last candle in snapshot: 13:30
2. immediately sent after connect: candles for 13:35 and 13:40
3. current date time: 13:41

this issue leads to an issue where, when with those 1-3 historic
candles a ma cross happens, it is executed - but too late.

to fix this, we get the current server time on start of the algo,
and then check if the incoming data is older.

related PRs:

#128
bitfinexcom/bfx-api-node-core#18

### DIFF
--- a/lib/ma_crossover/events/data_managed_candles.js
+++ b/lib/ma_crossover/events/data_managed_candles.js
@@ -21,7 +21,7 @@ const _reverse = require('lodash/reverse')
  */
 const onDataManagedCandles = async (instance = {}, candles, meta) => {
   const { state = {}, h = {} } = instance
-  const { args = {}, longIndicator, shortIndicator } = state
+  const { args = {}, longIndicator, shortIndicator, ts } = state
   const { symbol, long, short } = args
   const { debug, updateState, emitSelf, emit } = h
   const { chanFilter } = meta
@@ -36,6 +36,8 @@ const onDataManagedCandles = async (instance = {}, candles, meta) => {
 
   let indicatorsUpdated = false
   const [lastCandle] = candles
+
+  const outdated = ts > lastCandle.mts
 
   if (chanTF === long.candleTimeFrame) {
     indicatorsUpdated = true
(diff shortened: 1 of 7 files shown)


lib/ma_crossover/events/data_managed_candles.js
```diff
@@ -89,6 +91,11 @@ const onDataManagedCandles = async (instance = {}, candles, meta) => {
     await updateState(instance, { lastCandleShort: lastCandle })
   }
 
+  if (outdated) {
```
avsek477, line 94: I think if it's outdated, we should not update the candles at all in the state.

avsek477: Let me know If i understand it correctly and the following logic makes sense? Lets say `sendPing` returned `ts` as 10:03 am and we are subscribed to 5m candle, if i get the candle for 10:00 am we reject it since outdated will be true. But when we receive candle for 10:05 am, we update the `ts` and the long and short candle if matched. Because when its 10:10 am, we still get candle values for 10:05 am and because we aren't updating `ts` right now .. we end up with candle of 10:05 am which is the outdated candle since its already 10:10 am.

robertkowalski: > I think if it's outdated, we should not update the candles at all in the state. 

well we can pre-feed the cache, but should not trade on the data. the more data points we have, the more accurate our algo is

robertkowalski: using historic values to get the current ema value is fine, just not trading on them

robertkowalski: https://www.investopedia.com/ask/answers/122314/what-exponential-moving-average-ema-formula-and-how-ema-calculated.asp

avsek477 (approved):
lgtm

PR #127, opened 2021-04-21
check for ping pong amount and prices where required

avsek477:
Checks the algo order arguments for ping and pong amounts and prices.


robertkowalski (approved):
lgtm

PR #128, opened 2021-04-21
use manual data-channel-subscriptions

robertkowalski:
the current automagic does automatically subscribe to required
channels like candles and trades before each `life_start` run.

in `life_start` initialisations are done for acc/dis and ma cross.

through the automatic subscribe of data channels, it can happen
that data event hooks are called before `life_start` finished.
in those cases the algo order crashes. one solution is to run all
subscriptions after `life_start`, but this does not fit for all
algo orders, twap requires data already present at the time
`life_start` finishes.

this PR makes the data subscriptions manual, so a developer has to
decide for each algo order what the best behaviour is. in the
future we can make the subscriptions optionally awaitable so we
can make sure that subscriptions happened reliably at the current
point in time, which supports with developing better algo orders.

### DIFF
--- a/lib/twap/events/life_start.js
+++ b/lib/twap/events/life_start.js
@@ -16,7 +16,7 @@ const onLifeStart = async (instance = {}) => {
   const { state = {}, h = {} } = instance
   const { args = {} } = state
   const { priceTarget, priceCondition, orderType } = args
-  const { debug, emitSelf } = h
+  const { debug, emitSelf, subscribeDataChannels } = h
 
   if (!/MARKET/.test(orderType)) {
     if (_isFinite(priceTarget) && _isString(priceCondition)) {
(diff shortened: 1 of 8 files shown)


avsek477 (approved):
lgtm

vitaliystoliarovcc:
Is this WIP? Or will go into `master` with several TODOs?

robertkowalski:
> Is this WIP? Or will go into `master` with several TODOs?

the todos describe a feature thats going to be added soon, awaitable channel subscriptions:

>in the
future we can make the subscriptions optionally awaitable so we
can make sure that subscriptions happened reliably at the current
point in time, which supports with developing better algo orders.

lib/twap/events/life_start.js
```diff
@@ -29,6 +29,10 @@ const onLifeStart = async (instance = {}) => {
     }
   }
 
+  // TODO: use await here once awaitable feature is added
+  // so first tick can operate on data already
+  subscribeDataChannels(state)
```
vitaliystoliarovcc, line 34: lgtm. 
It seems that in this place, despite the issue with the lack of await, it behaves like the implementation in the master branch

vitaliystoliarovcc (approved):

PR #130, opened 2021-04-23
issue fix for invalid ping pong price if orderCount 1

avsek477:
Removes non-zero value check for ping pong prices if orderCount is greater than 1.

Task: https://app.asana.com/0/1125859137800433/1200244928837995


vitaliystoliarovcc (approved):
lgtm

PR #131, opened 2021-04-23
listen to correct event to start strategy

tarcisiozf:
In the example provided the current auth:success event is never called, we can use the `ready` instead

### DIFF
--- a/examples/ao_host.js
+++ b/examples/ao_host.js
@@ -59,7 +59,7 @@ host.on('ao:persist:db:update', async (updateOpts) => {
   debug('ao instance updated %s', updateOpts.gid)
 })
 
-host.on('ws2:auth:error', (packet) => {
+host.on('auth:error', (packet) => {
   debug('error authenticating: %j', packet)
 })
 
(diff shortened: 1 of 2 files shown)


examples/ao_host.js
```diff
@@ -67,7 +67,7 @@ host.on('error', (err) => {
   debug('error: %s', err)
 })
 
-host.once('ws2:auth:success', async () => {
+host.once('ready', async () => {
```
avsek477, line 70: there is one event `auth:success` will have to check if there is `ws2:auth:success`

avsek477: but yeah .. we aren't starting starting AO until we get `ready`

avsek477 (changes requested):
can you please make changes in readme as well :)

avsek477 (changes requested):
the same might not work for `ws2:auth:error` as well  .. it emits `auth:error` :)

avsek477 (approved):
lgtm

vitaliystoliarovcc (approved):
lgtm

PR #133, opened 2021-04-23
awaitable channel subscription

tarcisiozf:
ASANA Ticket: [data-channels: make them awaitable](https://app.asana.com/0/1200227858408358/1200221691555201)

to guarantee channels are subscribed when continuing with the algo

added as a bonus an option to not await the subscribe (fire & forget mode)


lib/host/gen_helpers.js
```diff
@@ -146,15 +172,43 @@ const genHelpers = (state = {}, adapter) => {
...
-      })
+          adapter.subscribe(connection, channel, payload)
+
+          if (options.fireAndForget) {
+            return Promise.resolve()
+          }
+
+          const { promise, resolve, reject } = flatPromise()
+          const timeout = options.timeout &&
+            setTimeout(() => reject(new Error(`Subscription to channel '${channel}' timed out`)), options.timeout)
```
avsek477, line 198: may be delete `pendingSubscriptionResponses[eid]` on timeout function as well? Otherwise, `pendingSubscriptionResponses` will have the data there always if not subscribed.

lib/host/gen_helpers.js
```diff
@@ -146,15 +172,43 @@ const genHelpers = (state = {}, adapter) => {
...
+          const eid = Math.random()
+
+          pendingSubscriptionResponses[eid] = {
```
avsek477, line 201: Also, can you create a function here to clear all timeouts of pendingSubscriptions if they exists so that we can call it on `life_stop` event of an algo order?

lib/host/gen_helpers.js
```diff
@@ -146,15 +172,43 @@ const genHelpers = (state = {}, adapter) => {
...
+      return Promise.all(
+        channels.map(({ channel, filter: payload }) => {
+          debug('subscribing to channel %j (filter: %) [AO gid %d]', channel, payload, gid)
 
-        // TODO: optional await
-        adapter.subscribe(connection, ch.channel, ch.filter)
-      })
+          adapter.subscribe(connection, channel, payload)
+
+          if (options.fireAndForget) {
```
avsek477, line 192: better to have this variable default to true if not provided .. i think

lib/host/gen_helpers.js
tarcisiozf: done!

lib/host/gen_helpers.js
tarcisiozf: done!

avsek477 (approved):
lgtm

robertkowalski (approved):
lgtm

PR #135, opened 2021-04-29
invalid slice amount error check issue fix

avsek477:
In accumulate/distribute algo, the validation was done for the `amount` value and not the `sliceAmount`. This PR fixes that.


robertkowalski (approved):
lgtm

PR #138, opened 2021-05-04
debug added properly while subscribing to data channels

avsek477:
Adding the missing params in debug while logging the subscribing channel info.


tarcisiozf (approved):
lgtm

PR #139, opened 2021-05-04
Ma crossover improvement

avsek477:
- Short indicator args issue fix
- MA values logged on csv file
- Skip ma values calculation for outdated candles
- Updated Tests for MA Crossovers
- Integration test added for golden and death cross with real candle values 

Task : https://app.asana.com/0/1125859137800433/1200228278955634


lib/ma_crossover/util/check_header.js
```diff
@@ -0,0 +1,32 @@
+'use strict'
+
+const fs = require('fs')
+
+const hasHeaderInFile = (path, headerArr) => {
+  return new Promise((resolve) => {
+    const rs = fs.createReadStream(path, { encoding: 'utf-8' })
```
robertkowalski, line 7: what happens when you create two MA cross orders in parallel?

robertkowalski: if they can both append to the file descriptor, then values will be mixed up...

how about a folder `ma_cross_logs` where on file is created for each algo order? then we can also delete the code for checking the header which might have some differences on windows based systems

lib/ma_crossover/events/life_start.js
```diff
@@ -1,6 +1,12 @@
 'use strict'
 
+const fs = require('fs')
+const os = require('os')
 const HFI = require('bfx-hf-indicators')
+const checkHeader = require('../util/check_header')
+const dir = `${os.homedir()}/.honeyframework/ma_crossover.csv`
```
robertkowalski, line 7: so far the repo was a library that did not knew that the app has a settings directory... hmmm...

how about passing a logdir option to algo_host? also did you think of the hosted version?

lib/ma_crossover/events/life_start.js
```diff
@@ -1,6 +1,12 @@
...
+
+const HEADERS_FOR_CSV = ['GID', 'TYPE', 'DATETIME', 'PRICE', 'PRICE_TYPE', 'MA_TYPE', 'MA_VALUE']
```
robertkowalski, line 9: how about a more general approach, where you can start the HF with a flag like `ALGO_LOG=true` and it would create CSVs for all algo orders?

could be encapsulated in a general function and available by a event like `await emit('exec:log_csv', gid, [value1, value2 ...])`

and in cases where logging is disabled, the general function immediately returns...

robertkowalski: we could add the base for it in this PR with the logging for ma cross, and then add other algos as well in seperate follow up PRs

robertkowalski (changes requested):
main issues:

1. the logging should be behind a flag
2. log dir must be configurable
3. logging can be more general, accessbile via `emit`

lib/ma_crossover/util/check_header.js
avsek477: hmm yeah .. thought we could use the filter function in sheets or excel to separate values using gid .. 

but yeah . separate logs for each algo order sounds a lot better .. will change it

lib/util/fs_operations.js
```diff
@@ -0,0 +1,38 @@
...
+
+const writeToFile = (stream, values) => {
+  if (_isString(values)) {
+    stream.write(values + '\n')
+    return
+  }
+  values.forEach(row => {
+    stream.write(row.join(',') + '\n')
+  })
+}
```
vitaliystoliarovcc, line 32: this seems to be tightly coupled with the code in `log_algo_data.js`. See the comment on that file

vitaliystoliarovcc (changes requested):

lib/util/fs_operations.js
```diff
@@ -0,0 +1,38 @@
...
+    await Fs.access(path)
+    return true
+  } catch {
+    return false
+  }
+}
+
+const createDir = async (path) => {
+  try {
+    await Fs.mkdir(path)
```
robertkowalski, line 15: should be recursive?

lib/host/events/log_algo_data.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  const dirExists = await fileOps.pathExists(logDir)
+
+  if (!dirExists) {
+    debug('creating logs folder(%s) for %s algo', filePath, id)
+    try {
+      await fileOps.createDir(logDir)
+    } catch (err) {
+      return debug('error creating logs folder(%s) for %s algo %O', filePath, id, err)
```
robertkowalski, line 42: if this case happens, this code is never called:

`await updateState(instance, { wls: writeAlgoLogStream })`

so the next algo order will crash... 

should we abort in general with an error already here and send it to the UI?

lib/util/fs_operations.js
robertkowalski: `await Fs.mkdir(path, { recursive: true })`

avsek477: done

lib/host/events/log_algo_data.js
```diff
@@ -0,0 +1,58 @@
...
+  const exists = await fileOps.pathExists(filePath)
+
+  const writeAlgoLogStream = fs.createWriteStream(filePath, { flags: 'a' })
+
+  if (!exists && !_isEmpty(headersForLogFile)) {
+    debug('creating file and writing headers in file(%s)', filePath)
+    values = [headersForLogFile, ...values]
+  }
+
+  fileOps.writeToFile(writeAlgoLogStream, values)
```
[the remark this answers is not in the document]
avsek477: created new class CsvLogWriter

lib/host/events/log_algo_data.js
avsek477: it wont... kept the algo orders alive even when error was found during logging

lib/ao_host.js
```diff
@@ -606,7 +610,7 @@ class AOHost extends AsyncEventEmitter {
    */
   async onAOStop (instance = {}, opts = {}) {
     const { h } = instance
-    const { channels = [], gid, connection, ev } = instance.state
+    const { channels = [], gid, connection, ev, wls } = instance.state
```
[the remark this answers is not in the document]
avsek477: changed to `algoLogWriter`

lib/ma_crossover/events/data_managed_candles.js
```diff
@@ -56,45 +60,55 @@ const onDataManagedCandles = async (instance = {}, candles, meta) => {
         longIndicator.add(price)
       } else if (state.lastCandleLong.mts === lastCandle.mts) {
         longIndicator.update(price)
+      } else if (state.lastCandleLong.mts > lastCandle.mts) {
+        updateLastCandleLong = false // do nothing
```
robertkowalski, line 64: you don't want to log ignored candles, or better even log them, but flag as `ignored = 1` in the csv

its required for calcs

avsek477: added a new column `CANDLE_IGNORED`

lib/ma_crossover/events/data_managed_candles.js
```diff
@@ -56,45 +60,55 @@ const onDataManagedCandles = async (instance = {}, candles, meta) => {
         longIndicator.add(price)
       } else if (state.lastCandleLong.mts === lastCandle.mts) {
         longIndicator.update(price)
```
robertkowalski, line 62: do this:

1. create algo order
2. submit a few market orders, which will lead to candle updates
3. check csv, for every update to one entry a csv log entry is written:

```
long,Fri May 14 2021 18:24:00 GMT+0200 (Central European Summer Time),51162,close,EMA(6),51220.504795251836,0
short,Fri May 14 2021 18:24:00 GMT+0200 (Central European Summer Time),51162,close,EMA(3),51203.59108453605,0
long,Fri May 14 2021 18:24:00 GMT+0200 (Central European Summer Time),51162,close,EMA(6),51220.504795251836,0
short,Fri May 14 2021 18:24:00 GMT+0200 (Central European Summer Time),51162,close,EMA(3),51203.59108453605,0
```

is this intended @avsek477 - can you check with ivan?

robertkowalski: having updates in the log helps with debugging, as we have intermediate states there

robertkowalski: hmmmm

avsek477: yes, it's intended .. we log and calculate ma values for every candle we receive

robertkowalski (approved):
lgtm

PR #141, opened 2021-05-12
DRAFT - mimic bfx websocket api

tarcisiozf:
### Description
The idea is to have a realistic bfx ws api mock for integration testing.

### Design Spec
**Real WS behavior**: 

With some handy components like the WsMock abstraction, we can create real ws servers and compose its message communication patterns. 
Providing us realistic network communication testing, enforce async and connection handling.

**Pseudo-random data generation**:

This allows generating data to simulate market moves without the necessity to store it.
Also, by reusing the same seed as input, we can replay events to generate the output in a deterministic manner.

**Extensible components**:

The idea is to expand the BitfinexApiMock using event emitters and adapters easily as we are aiming to have low-coupling
(eg. data generators for channel subscriptions)

### How to test

I've used the real example provided in `examples/ao_host.js` to create a base API mock that actually works, so basically you can set the environment variable pointing to the API mock:

* API_KEY=api key
* API_SECRET=api secret
* WS_URL=ws://localhost:42000

start the API mock:
```bash
node test/integration/public_api_example.js
```

start the example:
```bash
node examples/ao_host.js
```

### Possible future improvements
- [ ] Replay events coming from files (CSV?)
- [ ] Extract generic helpers (ws server/connection, pseudo-random data generator) into a different repo for reuse (maybe open-source it)
- [ ] Add chaos factors for message delivery to test resilience and guarantees of message delivery

PR #142, opened 2021-05-17
Benchmark

vitaliystoliarovcc:
Asana ticket: https://app.asana.com/0/1125859137800433/1200253930650939/f

Integrated Clinic.js and added couple of benchmark tests: connecting AO host and submiting order

~~This PR is still a draft since the created AO host communicates with the real Bitfinex API, and this leads to a couple of problems:~~
- ~~excessive load on API~~
- ~~API limits number of connections per time interval (returns error 429)~~

~~Therefore, we need emulation of Bitfinex API, which is already being worked on.~~

Depends on https://github.com/bitfinexcom/bfx-hf-algo/pull/141


benchmarks/reports/submit_order/4948.clinic-doctor/4948.clinic-doctor-systeminfo
```diff
@@ -0,0 +1,28 @@
+{
+  "clock": {
+    "hrtime": [
```
robertkowalski, line 3: reports should be gitignored

benchmarks/utils/setup.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const restify = require('restify')
+
+function setupBenchmark (execute) {
+  const server = restify.createServer()
+
+  server.get('/', async (req, res, next) => {
```
robertkowalski, line 8: why we do need a http server?

benchmarks/fixtures/orders.js
```diff
@@ -0,0 +1,28 @@
+function waitOrderStop (host, gid) {
```
robertkowalski, line 1: use strict

benchmarks/fixtures/orders.js
```diff
@@ -0,0 +1,28 @@
...
+      const { state = {} } = instance
+
+      if (state.gid === gid) resolve()
+    })
+  })
+}
+
+async function performOrder (host) {
+  const gid = await host.startAO('bfx-iceberg', {
+    symbol: 'tBTCUSD',
```
robertkowalski, line 13: maybe better to use a paper trading currency?

benchmarks/basic.js
```diff
@@ -0,0 +1,9 @@
+const { setupBenchmark } = require('./utils/setup')
```
robertkowalski, line 1: use strict, also in other files

README.md
```diff
@@ -23,17 +23,17 @@ const {
   AOHost, PingPong, Iceberg, TWAP, AccumulateDistribute, MACrossover
 } = require('bfx-hf-algo')
 
-const HFDB = require('bfx-hf-models')	
-const HFDBLowDBAdapter = require('bfx-hf-models-adapter-lowdb')	
-const {
```
robertkowalski, line 28: interesting!

for a 40 seconds timeframe you can see how memory usage is slowly increasing, even with the current api limitations

Screenshot 2021-05-18 at 10 48 05

robertkowalski: Screenshot 2021-05-18 at 10 52 15

benchmarks/reports/submit_order/4948.clinic-doctor/4948.clinic-doctor-systeminfo
vitaliystoliarovcc: where should they be stored?

benchmarks/utils/setup.js
vitaliystoliarovcc: `autocannon` requires it

README.md
vitaliystoliarovcc: Since it returns error 429 in the current implementation, for some reason Garbage collector doesn't clear these error objects. The following code produces an unhandledPromiseRejection

```js
setInterval(async () => {
  const host = spawnHost()

  await connectHost(host)
  await destroyHost(host)
}, 100)
```

image

Even if we ignore them, some data still remain in memory

```js
setInterval(async () => {
  const host = spawnHost()

  try {
    await connectHost(host)
    await destroyHost(host)
  } catch (e) {}
}, 100)
```

image

robertkowalski:
```
this.adapter.on('meta:error', this.onMetaError.bind(this))
```

from reading the code after your findings, i think the adapter event
listeners are never cleared:

https://github.com/bitfinexcom/bfx-hf-algo/blob/e498b0a869271efab6fe8490da86134ecef5784e/lib/ao_host.js#L98-L123

README.md
vitaliystoliarovcc: After deleting all listeners (`this.m.on` in `bfx-hf-algo/lib/ws_adapter.js`, `adapter.on` in `bfx-hf-algo/lib/host/ws2/bind_bus.js`, etc)

image

vitaliystoliarovcc: For examples, `(closure)` section contains only the external module `ws` among all entries 

image

vitaliystoliarovcc: image

benchmarks/submit_order.js
```diff
@@ -0,0 +1,13 @@
...
+const { setupBenchmark } = require('./utils/setup')
+const { spawnHost, connectHost } = require('./fixtures/host')
+
+const { performOrder } = require('./fixtures/orders')
+const host = spawnHost()
+const connectPromise = connectHost(host)
+
+setupBenchmark(async () => {
+  await connectPromise
+  await performOrder(host)
```
tarcisiozf, line 12: just to confirm, the host will be implicitly closed when the process finishes or we have to add that?

vitaliystoliarovcc: The WS server will receive `close` event whenever its client's process has terminated, right?

README.md
vitaliystoliarovcc: At the current stage of research I came to this code which is used in `bfx-api-node-core` dependency and it leads to the same memory leak


```js
const initWSState = require('bfx-api-node-core/lib/ws2/init_state')
const { WS_URL } = require('bfx-api-node-core/lib/config')
const WebSocket = require('ws')

/// setInterval here
/////

  const {
    apiKey,
    apiSecret
  } = wsSettings

  let wsState = null

  try {
    wsState = initWSState({
      url: WS_URL,
      apiKey,
      apiSecret,
      plugins: []
    })

  } catch (e) {
    console.log(e)
  } finally {
    if (!wsState) return
    const { ws } = wsState

    if (ws.readyState !== WebSocket.CLOSED) {
      ws.close()
    }
  }
```

image

vitaliystoliarovcc: Simplifed example:

```js
setInterval(async () => {
  const ws = new WebSocket('wss://api.bitfinex.com/ws/')
  ws.on('error', () => {})
  if (ws.readyState !== WebSocket.CLOSED) {
    ws.close()
  }
  ws.terminate()
}, 30)
```

benchmarks/basic.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+
+const { setupBenchmark } = require('./utils/setup')
+const { spawnHost, connectHost, destroyHost } = require('./fixtures/host')
+
+setupBenchmark(async () => {
+  const host = spawnHost()
+
+  await connectHost(host)
+  await destroyHost(host)
```
tarcisiozf, line 10: no need to await it

benchmarks/fixtures/orders.js
```diff
@@ -0,0 +1,30 @@
...
+  return new Promise((resolve) => {
+    host.once('ao:stop', (instance) => {
+      const { state = {} } = instance
+
+      if (state.gid === gid) resolve()
+    })
+  })
+}
+
+async function performOrder (host) {
```
tarcisiozf, line 13: there is no need to use async here

benchmarks/submit_order.js
tarcisiozf: I think we are good, please disregard that

README.md
vitaliystoliarovcc: Created the issue https://github.com/websockets/ws/issues/1889

robertkowalski: while preparing food it came to my mind:

```js
ws.on('error', () => {})
```

a `removeListener` is missing...

vitaliystoliarovcc: If the event isn't bound to the global scope, the handler will be removed along with `ws` object

vitaliystoliarovcc:
`submit_order` 😱 
image


image

benchmarks/utils/api_mock.js
```diff
@@ -0,0 +1,51 @@
+const ApiMock = require('../../test/util/mimic/bitfinex_api_mock')
```
robertkowalski, line 1: use strict?

robertkowalski:
```
$ TARGET=submit_order npm run benchmark:run

> bfx-hf-algo@4.0.3 benchmark:run /Users/robert/bitfinex/bfx-hf-algo
> API_URL=ws://localhost:5555 clinic doctor --dest ./.benchmarks_reports/$TARGET --on-port 'autocannon -d 30 localhost:$PORT' -- node --inspect ./benchmarks/$TARGET.js

Debugger listening on ws://127.0.0.1:9229/8fb2920b-b598-4548-9823-90e3912fdf1e
For help, see: https://nodejs.org/en/docs/inspector
Running 30s test @ http://localhost:3000
10 connections

running [                    ] 0%(node:5849) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 ws2:event:subscribed listeners added to [Manager]. Use emitter.setMaxListeners() to increase limit
```

benchmarks/utils/api_mock.js
vitaliystoliarovcc: Does `standard` support `use strict` rule checking?

vitaliystoliarovcc:
@robertkowalski yeah, according to the report https://github.com/bitfinexcom/bfx-hf-algo/pull/142#issuecomment-849006065 the memory leak occurs. And at least one of the reasons is event listeners.

I think it should be considered as a separate task/PR

benchmarks/utils/api_mock.js
vitaliystoliarovcc: it looks like `standard` is no doesn't have a such rule and will not be added

PR #143, opened 2021-05-19
storage: apply changes to receive algo order data

robertkowalski:
to support worker threads and in general better support for
hosted environments with many clients, storage won't be done
every few seconds any more, like currently the case

on start the state is stored, and on shutdown. this reduces the
db writes a lot (a twap with a 2 secod interval currently writes
every 2 seconds). it also brings flexibility by loosening the
coupling to the hf-models and database adapters, which we won't
use for the hosted version.

the event `ao:persist:db:update` is kept and can get used for
detailed UI updates on algo order details in the future

### DIFF
--- a/lib/ao_host.js
+++ b/lib/ao_host.js
@@ -89,9 +89,8 @@ class AOHost extends AsyncEventEmitter {
 
     this.ready = false
 
-    this.onAOStart = this.onAOStart.bind(this)
-    this.onAOStop = this.onAOStop.bind(this)
     this.onAOPersist = this.onAOPersist.bind(this)
+
     this.loadAO = this.loadAO.bind(this)
     this.triggerAOEvent = this.triggerAOEvent.bind(this)
     this.triggerGlobalEvent = this.triggerGlobalEvent.bind(this)
@@ -108,8 +107,6 @@ class AOHost extends AsyncEventEmitter {
     this.adapter.on('meta:reload', this.onMetaReload.bind(this))
     this.adapter.on('meta:connection:update', this.onMetaConnectionUpdate.bind(this))
 
-    this.on('ao:start', this.onAOStart)
-    this.on('ao:stop', this.onAOStop)
     this.on('ao:persist', this.onAOPersist)
 
     bindWS2Bus(this)
@@ -439,10 +436,8 @@ class AOHost extends AsyncEventEmitter {
    *
    * @param {string} id - algo order definition ID, i.e. bfx-iceberg
    * @param {object} args - algo order arguments/parameters
-   * @param {Function} [gidCB] - callback to acquire GID prior to ao:start
-   * @returns {string} gid - instance GID
    */
-  async startAO (id, args = {}, gidCB) {
+  async startAO (id, args = {}) {
     const ao = this.getAO(id)
 
     if (!ao) {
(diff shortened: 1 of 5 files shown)


lib/ao_host.js
```diff
@@ -501,13 +494,8 @@ class AOHost extends AsyncEventEmitter {
...
     await this.maybeCancelExistingOrders(state, gid)
 
-    if (_isFunction(gidCB)) {
-      await gidCB(gid)
-    }
-
-    await this.emit('ao:start', this.instances[gid])
-
-    return gid
+    const data = await this._startAlgo(this.instances[gid])
```
avsek477, line 497: similar operation for stopAO missing on line: 
https://github.com/bitfinexcom/bfx-hf-algo/blob/7bdc1e5af2145619625979b9931d94c0f79d3af9/lib/ao_host.js#L543

```
const data = await this._stopAlgo(this.instances[gid])
return data
```

robertkowalski: good catch! fixing

avsek477 (approved):
lgtm

lib/ao_host.js
avsek477: @robertkowalski this suddenly popped in my head haha.. this will be a problem

 https://github.com/bitfinexcom/bfx-hf-algo/blob/2321d1faa89b663a2611e3a7ed59c49ce771761e/lib/host/events/stop.js#L29

lib/ao_host.js
```diff
@@ -577,18 +572,15 @@ class AOHost extends AsyncEventEmitter {
...
 
     await withAOUpdate(this, gid, (instance = {}) => {
       const { state = {} } = instance
+      state.active = true
 
-      return {
-        ...state,
-        active: true
-      }
+      return false
```
vitaliystoliarovcc, line 583: `withAOUpdate` is useless since the callback always returns false.
Can be replaced with `instance.state.active = true`?

lib/ao_host.js
```diff
@@ -608,19 +604,20 @@ class AOHost extends AsyncEventEmitter {
...
 
     await withAOUpdate(this, gid, (instance = {}) => {
       const { state = {} } = instance
+      state.active = false
 
-      return {
-        ...state,
-        active: false
-      }
+      return false
```
vitaliystoliarovcc, line 620: the same as in the comment above

lib/ao_host.js
```diff
@@ -661,30 +663,24 @@ class AOHost extends AsyncEventEmitter {
...
-
     const { meta = {} } = ao
     const { serialize } = meta
 
-    if (!serialize) {
-      debug('can\t save AO %s [%s] due to missing serialize method', gid, id)
-      return
-    }
-
-    await this.emit('ao:persist:db:update', {
```
vitaliystoliarovcc, line 680: `ao:persist:db:update` also needs to be removed from README.md

vitaliystoliarovcc (changes requested):

lib/ao_host.js
robertkowalski: good point, fixed!

lib/ao_host.js
robertkowalski: fixed!

vitaliystoliarovcc (approved):
lgtm

PR #145, opened 2021-05-21
Indicator class not a constructor issue fix

avsek477:
The issue is originated while fetching the MA indicator class when the relativeCap or relativeOffset type is selected as Moving Average(MA). And since we have `SMA`  indicator available, acc/dist algo throws an error saying `Indicator class is not a constructor` resulting in failure to start the algo.

This fix here is by explicitly fetching the SMA or EMA indicator since `hasIndicatorOffset` and `hasIndicatorCap` only returns true if the given type is either 'ma' or 'ema'.

Task: https://app.asana.com/0/1125859137800433/1200358390931756

### DIFF
--- a/lib/accumulate_distribute/events/life_start.js
+++ b/lib/accumulate_distribute/events/life_start.js
@@ -44,7 +44,7 @@ const onLifeStart = async (instance = {}) => {
   }
 
   if (hasIndicatorCap(args)) {
-    const IndicatorClass = HFI[relativeOffset.type === 'ma' ? 'SMA' : 'EMA']
+    const IndicatorClass = HFI[relativeCap.type === 'ma' ? 'SMA' : 'EMA']
     const capIndicator = new IndicatorClass(relativeCap.args)
 
     debug('initialized cap indicator %s %j', relativeCap.type, relativeCap.args)
--- a/lib/accumulate_distribute/events/life_start.js
+++ b/lib/accumulate_distribute/events/life_start.js
@@ -32,7 +32,7 @@ const onLifeStart = async (instance = {}) => {
   )
 
   if (hasIndicatorOffset(args)) {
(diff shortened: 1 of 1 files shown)


vitaliystoliarovcc (approved):
lgtm

lib/accumulate_distribute/events/life_start.js
```diff
@@ -44,7 +44,7 @@ const onLifeStart = async (instance = {}) => {
   }
 
   if (hasIndicatorCap(args)) {
-    const IndicatorClass = HFI[relativeCap.type.toUpperCase()]
+    const IndicatorClass = HFI[relativeOffset.type === 'ma' ? 'SMA' : 'EMA']
```
tarcisiozf, line 47: is `relativeOffset` the correct variable in this case? cuz the if stmt is checking for indicator cap

lib/accumulate_distribute/events/life_start.js
```diff
@@ -32,7 +32,7 @@ const onLifeStart = async (instance = {}) => {
   )
 
   if (hasIndicatorOffset(args)) {
-    const IndicatorClass = HFI[relativeOffset.type.toUpperCase()]
+    const IndicatorClass = HFI[relativeOffset.type === 'ma' ? 'SMA' : 'EMA']
```
tarcisiozf, line 35: is it possible to index HFI by the final type instead of doing this check?

lib/accumulate_distribute/events/life_start.js
avsek477: no it's not .. fixed the typo.. thanks for pointing it out

lib/accumulate_distribute/events/life_start.js
avsek477: need to check for its dependency through out the algo .. will change it.. thanks!!

avsek477: Yes.. closing

PR #146, opened 2021-05-24
Acc dist moving averge typo and issue fixes

avsek477:
Since the error exists in acc/dist algo saying `Indicator class not a constructor` whenever moving average is selected by the user because of a typo, the following changes have been implemented along with some improvements. 

This PR includes two changes: 
1. Variable name and value changed from 'ma' to 'sma'.
2. updating the offset and relative indicator based on candle timeframe as selected by the user instead of adding all values to the indicator. 

Task: https://app.asana.com/0/1125859137800433/1200358390931756

### DIFF
--- a/lib/accumulate_distribute/events/data_managed_candles.js
+++ b/lib/accumulate_distribute/events/data_managed_candles.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const _reverse = require('lodash/reverse')
+const _last = require('lodash/last')
 const hasIndicatorOffset = require('../util/has_indicator_offset')
 const hasIndicatorCap = require('../util/has_indicator_cap')
 const parseChannelKey = require('../../util/parse_channel_key')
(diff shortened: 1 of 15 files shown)


vitaliystoliarovcc:
After submitting Acc/Dis order, the client application crashes (ui - master, server - master, algo - acc-dist-ma-issue-fixes)


image

On bfx-hf-algo/master it works correctly

avsek477:
> After submitting Acc/Dis order, the client application crashes (ui - master, server - master, algo - acc-dist-ma-issue-fixes)
> 
> image
> 
> On bfx-hf-algo/master it works correctly

rebased and pushed .. should work fine now

vitaliystoliarovcc (approved):
lgtm

robertkowalski (approved):
lgtm, can we add the logging system in the next step for the relative values?

avsek477:
> lgtm, can we add the logging system in the next step for the relative values?

will do

PR #147, opened 2021-05-24
Acc/DIst orderform validation

avsek477:
This PR adds the functionality to be able to fetch the error fields as well so that it's easier in the UI side to show the error message on the specific field of the algo order. It can be used by UI to check for errors before submitting the order to algo server. However, the UI if wishes to use this function, it must process params first using processParams function before using the validateParams function.

Task: https://app.asana.com/0/1125859137800433/1200373978248379

Related PR: https://github.com/bitfinexcom/bfx-hf-algo/pull/155

### DIFF
--- a/lib/accumulate_distribute/meta/validate_params.js
+++ b/lib/accumulate_distribute/meta/validate_params.js
@@ -5,6 +5,7 @@ const _isFinite = require('lodash/isFinite')
 const _isObject = require('lodash/isObject')
 const _isBoolean = require('lodash/isBoolean')
 const _includes = require('lodash/includes')
+const validationErrObj = require('../../util/validate_params_err')
 
 const ORDER_TYPES = ['MARKET', 'LIMIT', 'RELATIVE']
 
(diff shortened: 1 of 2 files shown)


vitaliystoliarovcc:
Depends on https://github.com/bitfinexcom/bfx-hf-algo/pull/155, right?

avsek477:
> Depends on #155, right?

No. However, in https://github.com/bitfinexcom/bfx-hf-algo/pull/155, while producing order amounts.. if total amount and sliced order amount entered by the user is less than the minimum order size, it results an empty array. So, this PR helps in avoiding such cases since we are checking valdiations for min and max order sizes here.

vitaliystoliarovcc:
Why not? This PR uses `pairConfig` parameter, which was introduced in https://github.com/bitfinexcom/bfx-hf-algo/pull/155

robertkowalski:
> Why not? This PR uses `pairConfig` parameter, which was introduced in #155

pairconfig was added with https://github.com/bitfinexcom/bfx-hf-algo/pull/120 and https://github.com/bitfinexcom/bfx-hf-server/pull/96

robertkowalski (approved):
lgtm

PR #152, opened 2021-05-25
emit algo order stop event required for ui updates

avsek477:
Emits `ao:stopped` event whenever algo order is stopped so that UI can be updated accordingly.


robertkowalski (approved):
lgtm

PR #153, opened 2021-05-25
return serialized data when ao is stopped

avsek477:
Returns the serialized data when ao is stopped so that the db can be updated.


robertkowalski (approved):
lgtm

PR #161, opened 2021-06-03
slice amount greater check issue fix for sell orders

avsek477:
Fixes the issue for slice amount being greater than total amount for sell orders.


robertkowalski:
thanks!

PR #164, opened 2021-06-09
skip outdated ma values for order price calculation

avsek477:
- Skips the ma calculated values from outdated candles


tarcisiozf (approved):
I'm guessing that we don't have good test coverage for this so this why there are no test changes.

vitaliystoliarovcc (approved):
lgtm

tarcisiozf (approved):
lgtm

robertkowalski (approved):
lgtm

PR #165, opened 2021-06-10
Improve integration tests

vitaliystoliarovcc:
- added tests for OCOCO, Ping/Pong, Acc/Dist and Iceberg ago orders
- fixed iteration of incoming/outgoing events in SessionSpy
- reduced interval of streaming to channel
- added Trades data provider
- fixed Book data provider
- extended BitfinexSessionMock to allow emulate outgoing events from the tests


test/integration/accumulate_distribute.js
```diff
@@ -0,0 +1,77 @@
...
+    spyConn
+      .received('subscribe', (e) => expect(e.channel).to.eq('book'))
+      .sent('subscribed', (e) => {
+        expect(e.channel).to.eq('book')
+        expect(e.chanId).to.be.a('number')
+      })
+      .received(NEW_ORDER, ({ fields: [placeholder, details] }) => {
+        expect(details.symbol).to.eq('tAAABBB')
+        expect(details.type).to.eq('EXCHANGE LIMIT')
+        expect(+details.amount).to.be.within(-0.2, -Number.EPSILON)
```
avsek477, line 72: makes more accurate if the range is between -1.2 to -0.8 according to the given input

avsek477 (approved):
rest lgtm

tarcisiozf (approved):
looks great

test/util/base_api_mock.js
```diff
@@ -39,11 +39,20 @@ class BaseApiMock extends ApiMock {
       low: 9_000,
       volume: 3
     })
+    const baseTrade = new PublicTrade({ price: 5000, amount: 9001 })
+    const baseBook = new OrderBook([
+      [67536333723, 500, -2],
+      [67538314922, 600, -2],
+      [67535007022, 500, 2],
```
robertkowalski, line 46: both heads at `500`, wouldn't they have matched?

robertkowalski: should be `499` for the buy instead?

test/util/mimic/data-providers/trades.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+// const timeFrames = require('bfx-hf-util/lib/candles/time_frames')
+// const parseChannelKey = require('../../../../lib/util/parse_channel_key')
```
robertkowalski, line 4: commented? should we remove?

test/util/mimic/responses/order.js
```diff
@@ -0,0 +1,40 @@
+module.exports = function ({ id, gid, cid, amount, status }) {
```
robertkowalski, line 1: use strict

test/util/base_api_mock.js
vitaliystoliarovcc: Yeah, it can, but it seems that in this situation it does not affect

robertkowalski (approved):
lgtm, will merge for 3.14

PR #169, opened 2021-06-22
forward auth token to ws manager

tarcisiozf:
The Ws Manager provided by the node-core lib already supports the usage of auth tokens as a replacement of API credentials, but currently, it does not sends such field for the manager instance.


robertkowalski (approved):
lgtm

PR #170, opened 2021-06-24
pass auth token in adapter

avsek477:
Check auth token as argument for manager in ws adapter in tests.


tarcisiozf (approved):
lgtm

robertkowalski:
lgtm

PR #172, opened 2021-06-25
relative type typo fix for logging in acc/dist algo

avsek477:
Typo fix for relative cap and offset order type.


robertkowalski (approved):
lgtm

PR #174, opened 2021-06-28
Add pair config to aohost

avsek477:
- Config pair added to AO host that can used to save in the instance state.
- Using pair config to generate allowed order sizes for the selected pair
- Util function to generate random number within the given range

### DIFF
--- a/lib/util/get_min_max_distorted_amount.js
+++ b/lib/util/get_min_max_distorted_amount.js
@@ -1,0 +1,29 @@
+'use strict'
+
+const _isFinite = require('lodash/isFinite')
+const { prepareAmount } = require('bfx-api-node-util')
+
+module.exports = (args = {}, pairConfig = {}) => {
+  const { amount, sliceAmount, amountDistortion = 0 } = args
+  const { minSize, maxSize } = pairConfig
+
+  const m = amount < 0 ? -1 : 1
+
+  let maxDistortedAmount = (1 + amountDistortion) * sliceAmount
+  let minDistortedAmount = (1 - amountDistortion) * sliceAmount
+
+  if (_isFinite(minSize)) {
+    if (Math.abs(minDistortedAmount) < minSize) minDistortedAmount = m * minSize
+    if (Math.abs(maxDistortedAmount) < minSize) maxDistortedAmount = m * minSize
+  }
+
+  if (_isFinite(maxSize)) {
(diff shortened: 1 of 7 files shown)


test/lib/util/get_min_max_distorted_amount.js
```diff
@@ -0,0 +1,36 @@
...
+
+  it('should return distorted amounts equal to the maximum order size', () => {
+    const orderParams = { amount: -200, sliceAmount: -150, amountDistortion: 0.25 } // sell order
+    const { minDistortedAmount, maxDistortedAmount } = getMinMaxDistortedAmount(orderParams, pairConfig)
+
+    assert.deepStrictEqual(minDistortedAmount, -100, 'should have equalled to maximum order size')
+    assert.deepStrictEqual(maxDistortedAmount, -100, 'should have equalled to maximum order size')
+  })
+
+  it('returns the distorted amounts properly', () => {
```
robertkowalski, line 29: can you add a test for floats?

avsek477: added.

lib/util/get_min_max_distorted_amount.js
```diff
@@ -0,0 +1,29 @@
...
+const { prepareAmount } = require('bfx-api-node-util')
+
+module.exports = (args = {}, pairConfig = {}) => {
+  const { amount, sliceAmount, amountDistortion = 0 } = args
+  const { minSize, maxSize } = pairConfig
+
+  const m = amount < 0 ? -1 : 1
+
+  let maxDistortedAmount = (1 + amountDistortion) * sliceAmount
+  let minDistortedAmount = (1 - amountDistortion) * sliceAmount
```
robertkowalski, line 13: try this input as a test:

```js
  it('should return distorted amounts equal to the maximum order size', () => {
    const orderParams = { amount: -0.3, sliceAmount: -0.1, amountDistortion: 0.25 } // sell order
    const { minDistortedAmount, maxDistortedAmount } = getMinMaxDistortedAmount(orderParams, pairConfig)

  })
```

robertkowalski: best way is to go via bignumber.js instead of applying `+prepareAmount` to the end result. in this was number values are correct at all times

vitaliystoliarovcc (approved):
lgtm

robertkowalski (approved):
lgtm

PR #176, opened 2021-07-12
Twap last min order issue fix

avsek477:
Issue: The last order generated may be less than the minimum order size because of the remaining amount, hence, causing it to stop the  algo order.

This PR provides fix to this issue.

### DIFF
--- a/lib/twap/meta/get_ui_def.js
+++ b/lib/twap/meta/get_ui_def.js
@@ -20,7 +20,7 @@ const getUIDef = () => ({
     'price targets, the price condition may be specified to match against',
     'either the order book or last trade. If a price delta is specified, the',
     'target must be within the delta in order to match.',
-    '\n\nNote: If amount distortion is provided, the total sliced order amounts will',
+    '\n\nNote: If amount distortion is provided, the total order amounts will',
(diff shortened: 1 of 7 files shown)


lib/twap/meta/get_ui_def.js
```diff
@@ -19,7 +19,9 @@ const getUIDef = () => ({
     'that must be conditionally matched against another factor.\n\nWith custom',
     'price targets, the price condition may be specified to match against',
     'either the order book or last trade. If a price delta is specified, the',
-    'target must be within the delta in order to match.'
+    'target must be within the delta in order to match.',
+    '\n\nNote: If amount distortion is provided, the total sliced order amounts will',
```
robertkowalski, line 23: `the total order amounts will`

robertkowalski (changes requested):
small text issue

also last order fails for me with: invalid order size: minimum amount required -- is that expected?

lib/twap/meta/get_ui_def.js
avsek477: fixed

robertkowalski:
lgtm

PR #178, opened 2021-07-14
use flake ids

tarcisiozf:
The actual key is a combination of gid with the name of the algo order, whereas gid is a timestamp. this did work in the desktop app, but for the hosted version we'll have conflicts at one point, so gid should become a global distributed ID.

Flake ID generator yields k-ordered, conflict-free ids in a distributed environment.

This way we can set the cluster id as datacenter and shard id as worker id.

### DIFF
--- a/lib/util/gen_client_id.js
+++ b/lib/util/gen_client_id.js
@@ -1,9 +1,20 @@
 'use strict'
 
-let last = Date.now()
+const FlakeId = require('flake-idgen')
+const format = require('biguint-format')
+
+let generator
 
 module.exports = () => {
-  const now = Date.now()
-  last = (last < now) ? now : last + 1
-  return last
+  if (!generator) {
+    const {
+      FLAKE_WORKER_ID: worker,
+      FLAKE_DATACENTER_ID: datacenter
+    } = process.env
+
+    generator = new FlakeId({ worker, datacenter })
+  }
+
+  const id = generator.next()
+  return format(id, 'dec')
 }
--- a/package.json
+++ b/package.json
@@ -47,8 +47,10 @@
     "bfx-api-node-util": "^1.0.2",
     "bfx-hf-indicators": "^2.0.8",
     "bfx-hf-util": "^1.0.1",
+    "biguint-format": "^1.0.2",
     "bluebird": "^3.5.5",
     "debug": "^4.3.1",
+    "flake-idgen": "^1.4.0",
     "flat-promise": "^1.0.2",
     "lodash": "^4.17.15"
   },


vigan-abd:
hey @tarcisiozf @robertkowalski maybe we could use something simple like this?
```js
Buffer.concat([
  Buffer.from([0x01]), // machine id [0x00-0xFF]
  crypto.randomBytes(31) // random id
]).readUInt32LE()
```

tarcisiozf:
> hey @tarcisiozf @robertkowalski maybe we could use something simple like this?
> 
> ```js
> Buffer.concat([
>   Buffer.from([0x01]), // machine id [0x00-0xFF]
>   crypto.randomBytes(31) // random id
> ]).readUInt32LE()
> ```

On large scale, like Twitter, Instagram and other random can be dangerous, even though the conflict probability is low. But for our case I agree, we can use a simpler approach.

vigan-abd:
> > hey @tarcisiozf @robertkowalski maybe we could use something simple like this?
> > ```js
> > Buffer.concat([
> >   Buffer.from([0x01]), // machine id [0x00-0xFF]
> >   crypto.randomBytes(31) // random id
> > ]).readUInt32LE()
> > ```
> 
> On large scale, like Twitter, Instagram and other random can be dangerous, even though the conflict probability is low. But for our case I agree, we can use a simpler approach.

well I just followed what's used on uuid/v4 lib: https://github.com/uuidjs/uuid/blob/1c849da6e164259e72e18636726345b13a7eddd6/src/rng.js#L6

the only diff is that they generate a large rand buffer (256) and continue to slice it by 16 bytes until it reaches the end and then they generate again the same buffer. We could actually do the same I think:
```js
// uuid/v4
import crypto from 'crypto';

const rnds8Pool = new Uint8Array(256); // # of random values to pre-allocate
let poolPtr = rnds8Pool.length;

export default function rng() {
  if (poolPtr > rnds8Pool.length - 16) {
    crypto.randomFillSync(rnds8Pool);
    poolPtr = 0;
  }
  return rnds8Pool.slice(poolPtr, (poolPtr += 16));
}
```

```js
const rng = require('uuid/src/rng')

const guid = (machineId) => Buffer.concat([
   Buffer.from([machineId]),
   rng.slice(1),
   rng()
]).readUInt32LE()
```

Since in our approach we use also this machineId we're sure that numbers are unique per each host/worker and we can spawn up to 256 workers, we could increase this to 512 if we want to be sure that we don't have scaling issues.

also since crypto.randomBytes uses `dev/urandom` we should be safe especially with sync call I guess

Anyway also snowflake id generator looks good to me, we can choose either method I guess

PR #181, opened 2021-08-16
dont stop algo for postonly cancelled cases

avsek477:
When an active algo receives postonly cancelled orders, it stops the running algo since it considers it as a cancelled order. This is a problem for postonly orders, so a fix is provided for this case.


vigan-abd (approved):
lgtm

PR #183, opened 2021-09-08
return empty array if none prices match in the ping or pong array table

avsek477:
Task: https://app.asana.com/0/1125859137800433/1200940826032403

- Return an empty array if the prices don't match from the pingPongTable or the activePongs array
- extract values from activePongs in case of pong order fill

### DIFF
--- a/lib/ping_pong/util/ping_pong_table.js
+++ b/lib/ping_pong/util/ping_pong_table.js
@@ -11,6 +11,11 @@ const append = (pingPongTable, pingPrice, pongPrice) => {
 
 const extract = (pingPongTable, price) => {
   const pingItem = pingPongTable.find(item => item[0] === price)
+
+  if (!pingItem) {
+    return []
+  }
+
   const pongPrice = pingItem[1]
   const nextPingPongTable = pingPongTable.filter(item => item !== pingItem)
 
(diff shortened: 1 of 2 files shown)


vitaliystoliarovcc (approved):
lgtm

pointops:
That works.

lib/ping_pong/util/ping_pong_table.js
```diff
@@ -11,6 +11,11 @@ const append = (pingPongTable, pingPrice, pongPrice) => {
 
 const extract = (pingPongTable, price) => {
   const pingItem = pingPongTable.find(item => item[0] === price)
```
tarcisiozf, line 13: what are the odds of `pingPongTable` being empty? `item[0]` would break it too.

vigan-abd:
```suggestion
  const pingItem = pingPongTable.find(item => item && item[0] === price)
```

avsek477: `pingPongTable` is a 2d array.. `pingPongTable` array being empty won't break the code .. and if there exists an item, it will always have the ping prices at index 0 and pong prices at index 1(handled this way by the `append` function).. so i think this check is unncessary

PR #184, opened 2021-09-09
Internationalization support

vitaliystoliarovcc:
- Add i18n for labels (`gen_order_label.js` modules, saves 'i18n' property into AO state)
- Add i18n for validation messages (`validate_params.js` modules return `object with `i18n` property)
- Add i18n for UIDef
example: 
```js
// returns the same schema as before, but with translated labels and titles
getUIDef({ i18n: { t: <translate function>, prefix: <string> }}) 
```

### DIFF
--- a/lib/util/i18n.js
+++ b/lib/util/i18n.js
@@ -1,7 +1,11 @@
 'use strict'
 
 function t (options, key, original, props) {
-  return options ? (options.t(key, props) || original) : original
+  if (!options) return original
+
+  const fullKey = [options.prefix, key].join('')
+
+  return options.t(fullKey, props) || original
 }
 
 function apply (object, i18nKey, i18nProps) {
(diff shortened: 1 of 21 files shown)


avsek477 (approved):
lgtm

lib/util/i18n.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+function t (options, key, original, props) {
```
vigan-abd, line 3: hey @vitaliystoliarovcc maybe we should rename options param to i18n? I think it's a bit confusing right now :sweat_smile:

vigan-abd (changes requested):
really good work, just a minor suggestion

lib/util/i18n.js
vitaliystoliarovcc: Actually options here are object `{t: function(){}, prefix: string}` 

I'll add the documentation 😅

vitaliystoliarovcc: done

vigan-abd (approved):

PR #189, opened 2021-10-07
Fix label compatibility with UI

ghost:
Asana ticket: https://app.asana.com/0/1125859137800433/1201143788684035

image

### Description

The UI at https://trading.bitfinex.com only accepts specific format in the order's `meta` field (where `label` is a string, but it isn't documented). At the same time the third-party services can submit any different format (HF in our case) and Bitfinex API accepts this (@prdn can API have contracts for validation of meta field?) 

### Fixes

- Prevent runtime error that happens due to a change in the format returned by genOrderLabel (for internationalization purposes). Now `label` field is a string again

### DIFF
--- a/lib/host/events/submit_all_orders.js
+++ b/lib/host/events/submit_all_orders.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const _isObject = require('lodash/isObject')
+const _isPlainObject = require('lodash/isPlainObject')
 const _omit = require('lodash/omit')
 const _isFunction = require('lodash/isFunction')
 const _isString = require('lodash/isString')
@@ -18,9 +18,7 @@ const withAOUpdate = require('../with_ao_update')
 module.exports = async (aoHost, gid, orders) => {
   const { instances } = aoHost
   const instance = instances[gid]
-  const _orders = _isObject(orders)
-    ? Object.values(orders)
-    : orders
+  const _orders = Object.values(orders)
 
   if (!instance) {
     return
@@ -63,7 +61,7 @@ module.exports = async (aoHost, gid, orders) => {
 
       if (_isString(label)) {
         order.meta.label = label
-      } else if (_isObject(label) && _isString(label.origin)) {
+      } else if (_isPlainObject(label) && _isString(label.origin)) {
         order.meta.label = label.origin
         order.meta.i18n = {
           ...(order.meta.i18n || {}),


lib/host/events/submit_all_orders.js
```diff
@@ -58,7 +59,17 @@ module.exports = async (aoHost, gid, orders) => {
     if (!order.meta) order.meta = {}
 
     if (_isFunction(genOrderLabel)) {
-      order.meta.label = genOrderLabel(state)
+      const label = genOrderLabel(state)
+
+      if (_isString(label)) {
+        order.meta.label = label
+      } else {
```
eugeneglova, line 66: I'd add `} else if (_isObject(label)) {` or `} else if (_has(label, 'origin')) {`

vitaliystoliarovcc: db491695f4aa027287e052ed74eef6052d1863b2

lib/host/events/submit_all_orders.js
```diff
@@ -3,6 +3,8 @@
 const _isObject = require('lodash/isObject')
 const _omit = require('lodash/omit')
 const _isFunction = require('lodash/isFunction')
+const _isString = require('lodash/isString')
+const isObject = require('lodash/isObject')
```
eugeneglova, line 7: for consistency maybe rename `isObject` to `_isObject`

vitaliystoliarovcc: fixed

lib/host/events/submit_all_orders.js
```diff
@@ -58,7 +59,19 @@ module.exports = async (aoHost, gid, orders) => {
     if (!order.meta) order.meta = {}
 
     if (_isFunction(genOrderLabel)) {
-      order.meta.label = genOrderLabel(state)
+      const label = genOrderLabel(state)
+
+      if (_isString(label)) {
+        order.meta.label = label
+      } else if (_isObject(label) && _isString(label.origin)) {
```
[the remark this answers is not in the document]
vitaliystoliarovcc: @vigan-abd only in this case or on line 21 too?

```js
  const _orders = _isObject(orders)
    ? Object.values(orders)
    : orders
```
and after it is used as array 
```js
 _orders.reduce((obj, ///
_orders.map(///
```

Probably the ternary operator is useless here, because `_isObject(orders)` always `true`, regardless of whether the parameter is an object or an array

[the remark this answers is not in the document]
vitaliystoliarovcc: fixed

PR #193, opened 2021-12-15
return algo id in start and stop function of aohost

avsek477:
Return algo id  on start and stop algo functions which is required to be sent by the ao events in `bfx-hf-server`.


vigan-abd (approved):
lgtm

vigan-abd (approved):
lgtm

PR #198, opened 2022-02-09
Trace AO signals

tarcisiozf:
We need to improve the analyses of how a given algo order is working under the hood, and we need to visualize better how the signals are handled and what a given event (like an order fill) can cause.

We introduced a signal tracer tool to log this data locally to solve that. The file later can be parsed and investigated.

The implementation is heavily inspired by the OpenTelemetry libraries, where data is represented by a graph of spans displayed on a timeline.

-----

Examples of visualization: https://www.loom.com/share/b8a4c5adf78149578a17138f92429b68

PS: the visualization tool is under development

### DIFF
--- a/lib/host/signals/local_storage.js
+++ b/lib/host/signals/local_storage.js
@@ -1,0 +1,54 @@
+'use strict'
+
+const { createWriteStream } = require('fs')
+const { stat, mkdir } = require('fs/promises')
+const path = require('path')
+const flatPromise = require('flat-promise')
+
+class LocalStorage {
+  constructor (dir, fileName) {
+    this.dir = dir
+    this.fileName = fileName
+    this.isOpen = false
+  }
+
+  async start () {
+    this.isOpen = true
+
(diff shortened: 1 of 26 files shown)


avsek477 (approved):
lgtm 👍

lib/host/signals/local_storage.js
```diff
@@ -0,0 +1,54 @@
+'use strict'
+
+const { createWriteStream } = require('fs')
+const { stat, mkdir } = require('fs/promises')
```
vigan-abd, line 4: @tarcisiozf @avsek477 this doesn't seem to be supported by node v12, see https://github.com/bitfinexcom/bfx-hf-ui-core/runs/5248400666?check_suite_focus=true
please use require('fs').promises to have backward compatibility on future

tarcisiozf: We were planning to upgrade to node 14 and we had to fallback, will fix

PR #199, opened 2022-02-18
fix local storage

tarcisiozf:
when this lib is imported by electron fs gets replaced by an empty mock, we should lazy load those functions


.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 12.22.8
```
vigan-abd, line 1: what's .tool-versions, is it necessary to be included in repo?

tarcisiozf: It's a common pattern for IDEs and version managers (like asdf) to detect the best version to use

tarcisiozf: The version not being specified helped the bug not to be detected

vigan-abd: will this have any effect in cli runtime?

vigan-abd (approved):

.tool-versions
tarcisiozf: it only affects tools like ASDF or IDEs

PR #213, opened 2022-09-15
Revert twap trade beyond end

avsek477:
TWAP algo should not be stopped in case of `tradeBeyondEnd` option.

Task: https://app.asana.com/0/1201849173362898/1202960193464620


arifjmamun (approved):
LGTM.

PR #224, opened 2023-02-27
use packetWDDelay from user setting

arifjmamun:
task: https://app.asana.com/0/1201849173362898/1203393471944924/f

### DIFF
--- a/lib/ws_adapter.js
+++ b/lib/ws_adapter.js
@@ -51,7 +51,8 @@ module.exports = class AOAdapter extends EventEmitter {
       dms,
       withHeartbeat,
       affiliateCode,
-      plugins = []
+      plugins = [],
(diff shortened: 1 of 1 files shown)


lib/ws_adapter.js
```diff
@@ -51,7 +51,8 @@ module.exports = class AOAdapter extends EventEmitter {
       dms,
       withHeartbeat,
       affiliateCode,
-      plugins = []
+      plugins = [],
+      packetWDDelay
```
avsek477, line 55: Please add a default value as well. You can use 30 seconds as default.

avsek477 (changes requested):

avsek477 (approved):

tarcisiozf (approved):

PR #228, opened 2023-08-14
Added function to restart ws connection.

tekwani:
Added function to restart ws connection.
Monitor last ws message received to check packet delay timeout for sending meta reload event.

### DIFF
--- a/lib/ao_host.js
+++ b/lib/ao_host.js
@@ -27,6 +27,8 @@ const bindWS2Bus = require('./host/ws2/bind_bus')
 const initAO = require('./host/init_ao')
 const initAOState = require('./host/init_ao_state')
 
+const { WD_PACKET_DELAY } = require('./constants')
+
 /**
  * @typedef {object} EventMetaInformation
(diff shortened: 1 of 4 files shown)


lib/ao_host.js
```diff
@@ -222,8 +238,12 @@ class AOHost extends AsyncEventEmitter {
   }
 
   onMetaReload () {
-    this.cleanState()
-    this.emit('meta:reload')
+    const sinceLastMessage = Date.now() - this.lastMessageTime
+    // check if packet delay timed out
+    if (sinceLastMessage > this.packetWDDelay) {
```
avsek477, line 243: Won't this result in algo orders being stopped and users not being given the option to resume/cancel them in certain cases?

avsek477 (commented):
rest, lgtm

lib/ao_host.js
tekwani: Fixed. Added packetWDDelay as ws timeout for dms control.

avsek477 (approved):

